PR #1, opened 2015-10-09
Adding proxy support for token_from_code and token_from_refresh methods

glukacsy:
Hi, 

I noticed that the token_from_code we are using in our project fails in an environment requiring proxy. The fundamental reason for this is that token_from_code under the hood instantiates a plain http_client without passing any configuration to it. 

I was considering several designs how to solve this, but for simplicity I settled down on the idea of setting the proxy directly on the Oauth2 object.

Let me know if you have a different design in mind - in general I felt that the OAuth2 config is somewhat detached from the encapsulating HTTP config object. For example, if the OAuth2 config object knew about the parent HTTP config object it would be able to pull the proxy (and other things if needed) from the parent. I was just not sure about the design intent here and didn't want to propose a bigger structural change.

Regards,
Gergely

### DIFF
--- a/Release/include/cpprest/oauth2.h
+++ b/Release/include/cpprest/oauth2.h
@@ -446,6 +446,24 @@ public:
     /// Default: "access_token".
     /// </summary>
     void set_access_token_key(utility::string_t access_token_key) { m_access_token_key = std::move(access_token_key); }
+	
+	/// <summary>
+	/// Get the web proxy object
+	/// </summary>
+	/// <returns>A reference to the web proxy object.</returns>
+	const web_proxy& proxy() const
+	{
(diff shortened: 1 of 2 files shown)


kavyako:
hey glukacsy

Sorry for our delayed response.
The changes look good. To contribute to the project, you will need to sign our [contributor license agreement (CLA)](https://www.codeplex.com/Download?ProjectName=casablanca&DownloadId=623578).  Once you've signed you can email it directly to me at kavyako at Microsoft dot com or askcasablanca at Microsoft dot com.

Thanks
Kavya.

glukacsy:
Hey, I have emailed you the CLA, thank you very much!

PR #15, opened 2015-11-10
Add unsolicited pong support

glukacsy:
Hi,

I added unsolicited pong support to Casablanca (see https://tools.ietf.org/html/rfc6455#section-5.5.3). This is useful for cases when the client wants to check whether the websocket connection is still up and running. For example, if I unplug my network cable, without this change, Casablanca simply does not notice that the Websocket connection is down, the close handler won't get invoked etc. 

With these changes, the client can send (periodically or when certain events happen, e.g. network change events) an unsolicited pong request which has no effect on a healthy websocket connection, but will help Casablanca/WebSocket++ detect that the connection has been lost. With these changes, if I unplug my network cable and if I send a pong message after that, Casablanca correctly invokes the websocket close handler within a few seconds.

Regards,
Gergely

### DIFF
--- a/Release/include/cpprest/ws_msg.h
+++ b/Release/include/cpprest/ws_msg.h
@@ -73,6 +73,15 @@ class websocket_outgoing_message
 {
 public:
 
+	/// <summary>
+	/// Sets a the outgoing message to be an unsolicited pong message.
+	/// This is useful when the client side wants to check whether the server is alive.
+	/// </summary>
+	void set_pong_message()
+	{
+		this->set_message_pong();
+	}
+
     /// <summary>
     /// Sets a UTF-8 message as the message body.
     /// </summary>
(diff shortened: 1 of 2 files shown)


kavyako:
hey Gergely

For any changes, especially that add new public APIs, we would really appreciate sufficient test coverage.
The websockets tests are under cpprestsdk / Release / tests / functional / websockets / client /  
Could you please take a look at adding a test if possible? 

Let me know if you have any concerns here.
 
Thanks
Kavya

glukacsy:
Hi, no problem at all, I will add tests ASAP (I should have done it with the original pull request). 

Thanks, 
Gergely

glukacsy:
Hi, I added a new pull request which is the same as this + integration tests. I guess we can close this one?

Regards,
Gergely

kavyako:
Closing this PR. 
Now being tracked under https://github.com/Microsoft/cpprestsdk/pull/20

Issue #13, opened 2015-11-09
Deadlock when closing the http_listener


gigaplex:
Duplicated from CodePlex (issue 421) since the documentation says that issues should be reported to GitHub instead:

When calling http_listener.close().wait() as we're shutting down our service, the application will occasionally deadlock because the wait operation never returns. The deadlock appears to be in the http_windows_server::stop() function, there's a wait that never returns.

    pplx::task<void> http_windows_server::stop()
    {
        // Shutdown request queue.
        if(m_hRequestQueue != nullptr)
        {
            HttpShutdownRequestQueue(m_hRequestQueue);
             m_receivingTask.wait();

            // Wait for all requests to be finished processing.
            m_zeroOutstandingRequests.wait(); // Deadlock appears to be here

            HttpCloseRequestQueue(m_hRequestQueue);
        }

gigaplex:
I suspect this area of code is of interest:

    pplx::task<void> http_windows_server::respond(http::http_response response)
    {
        windows_request_context * p_context = static_cast<windows_request_context *>(response._get_server_context());
        return pplx::create_task(p_context->m_response_completed).then([p_context](::pplx::task<void> t)
        {
            // After response is sent, break circular reference between http_response and the request context.
            // Otherwise http_listener::close() can hang.
            p_context->m_response._get_impl()->_set_server_context(nullptr);
            t.get();
        });
    }

It looks like if there's some exception that prevents this bit of code from running, then the circular reference isn't broken, causing the hang.

gigaplex:
From `windows_request_context::async_process_request`:    

    const unsigned long error_code = HttpReceiveHttpRequest(
            pServer->m_hRequestQueue,
            m_request_id,
            0,
            m_request,
            headers_size,
            NULL,
            &m_overlapped);

    if(error_code != NO_ERROR && error_code != ERROR_IO_PENDING)
    {
        CancelThreadpoolIo(pServer->m_threadpool_io);
        m_msg.reply(status_codes::InternalError);
    }

[...]

PR #16, opened 2015-11-12
Initial attempt at win32 http_listener response refactor

gigaplex:
For Issue 13: https://github.com/Microsoft/cpprestsdk/issues/13

Here is a first cut at a refactor for the response handling to reduce cyclic reference leaks and allow error replies to propagate back to the client. I'm not 100% happy with it, the content_ready_task for example will potentially race against async_process_response if they both attempt to call cancel_request under certain error conditions. There's also no attempt to catch exceptions on the proxy_content_ready completion event, which might cause some unobserved exception issues. I've created the pull request at this early stage to facilitate discussion.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -34,3 +34,6 @@ Cisco Systems
 Gergely Lukacsy (glukacsy)
 
 thomasschaub
+
+Trimble
+Tim Boundy (gigaplex)
--- a/Release/include/cpprest/details/http_server_httpsys.h
+++ b/Release/include/cpprest/details/http_server_httpsys.h
@@ -99,8 +99,13 @@ struct windows_request_context : http::details::_http_server_context
     // Dispatch request to the provided http_listener.
     void dispatch_request_to_listener(_In_ web::http::experimental::listener::details::http_listener_impl *pListener);
 
-    // Initialise the response task callbacks
-    void do_response(bool bad_request);
+    enum ShouldWaitForBody
+    {
+        WaitForBody,
+        DontWaitForBody
+    };
+    // Initialise the response task callbacks. If the body has been requested, we should wait for it to avoid race conditions.
+    void init_response_callbacks(ShouldWaitForBody shouldWait);
 
     // Read in a portion of the request body.
     void read_request_body_chunk();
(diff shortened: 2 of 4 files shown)


gigaplex:
CLA is awaiting signature from our legal team, they've informally granted permission at this stage.

ras0219-msft:
To improve readability, could you use an enum for the `bad_request` argument of `do_response` instead of a `bool`? The meaning of `do_response(response::GOOD_REQUEST)` is more obvious compared `do_response(false)`.

As mentioned in the other thread, I'm concerned about race conditions regarding the reading and setting of `m_msg`. If there are possible concurrent accesses, I think every reader will need to "pin" it to the stack by making a local copy of the `shared_ptr` so that it doesn't get deleted while they are inside a call. Example:

* To start, `m_msg` is the only strong ref to the actual request object.
* Thread 1 calls `m_msg.reply()` ([link](https://github.com/Microsoft/cpprestsdk/pull/16/files#diff-8a511de9f5257644c2da726766ae9a21R538)).
* After the pointer load completes on Thread 1, Thread 2 calls `m_msg = http_request()` ([link](https://github.com/Microsoft/cpprestsdk/pull/16/files#diff-8a511de9f5257644c2da726766ae9a21R714)). This removes the last remaining strong ref to the actual request object, so it is deleted.
* Thread 1 proceeds to modify the now-freed request object, invoking UB of the worst sort.

gigaplex:
Thanks for the initial feedback. I chose to use a bool instead of an enum because that's how the asio code currently does it, but I agree, I usually prefer enums here.

As for the race condition in your example, m_msg isn't the only strong ref to the request object. The content_ready_task object contains a strong reference too. It's copied in the http_request::content_ready task creation function. I have however found a race condition of a similar nature, which I'm working to address.

gigaplex:
I submitted the CLA a few days ago, yet this is still tagged as cla-required. How long does it normally take to process?

kavyako:
yes, this seems to be taking unusually long, I am following up on this with our CLA automation team.

gigaplex:
I'm fairly happy with the current state of my changes now. Can I get another round of review for merging please?

One question I do have is that in the `m_response_completed` callback, immediately after `p_context->m_response._get_impl()->_set_server_context(nullptr);` there was a call to `t.get();`. What is the purpose of this call? It wasn't carried over in my refactor, but if it's needed for a specific reason, I can add it back in.

gigaplex:
I've rebased against 2.7.0.

ras0219-msft:
Sorry for the delay on this one, thank you very much for your patience and the contribution!

PR #26, opened 2015-11-27
Static lib compilation for UWP

simonferquel:
This is a very small addition with only a vcxproj for doing static lib compilation of Casablanca for UWP platform.
This produces safe to ignore warnings about static lib compilation with /ZW flag (it is safe because Casablanca does not expose public winrt types).

The motivation was to mix the benefits of using the native Windows http client (with its support for caching, prefetch and more), and the completeness of Casablanca Json serialization/deserialization (WinRT Json is somewhat basic, especially for long numbers handling), without paying for the whole Casablanca dll (saving bandwidth and memory consumption - especially in background agents).


Release/src/build/vs14.uwp/cpprestsdk140.uwp.staticlib.vcxproj
```diff
@@ -0,0 +1,110 @@
...
+    <Import Project="..\winrt.vcxitems" Label="Shared" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+    <DebugFileSuffix>d</DebugFileSuffix>
+  </PropertyGroup>
+  <PropertyGroup>
+    <TargetName>$(CppRestBaseFileName)140$(DebugFileSuffix)_uwp_$(CppRestSDKVersionFileSuffix)</TargetName>
```
kavyako: The import lib for the shared library cpprest140d_uwp_2_7.dll already has this name. I would prefer using different names here to minimize chances of mistaking the library type (import/static).
How about following a specific naming convention for static libs, say add _static suffix (cpprest140d_uwp_static_2_7.lib) or prepend the name with lib (libcpprest140d_uwp_2_7.lib).

kavyako: Some one suggested we can look at doing something similar to boost, http://www.boost.org/doc/libs/1_42_0/more/getting_started/windows.html#library-naming

If you are building a static library, use the lib prefix.
If the library is linking statically to the C++ standard library and compiler runtime support libraries, use s in the .lib name.
In this case, we are not linking statically to the runtime. so we can go with libcpprest140d_uwp_2_7.lib.
In future, if someone wants to add another version to link to the runtime statically, we can add _s to the name.
What do you think.

kavyako:
This change got merged to the development branch while I was merging something else. I will not revert it. Once we decide on the name, I can update the proj file with it.

kavyako:
I have modified the static lib name to libcpprest140d_uwp_2_7.lib. The change is in the development branch. Please reopen the thread if you disagree with the naming convention.

Thank you for contributing the changes.

PR #38, opened 2015-12-21
Unknown typename web_proxy in oauth2_h

georgemp:
When compiling on OS X and using OAuth2, we get an error that web_proxy is an unknown type. Including cpprest/details/web_utilities.h to bring in the definition for web_proxy


kavyako:
Thank you for fixing the issue. This has been merged to development branch.
We try to follow a consistent coding style, I have fixed the indentation in the oauth2.h here in my commit.

Issue #9, opened 2015-10-29
Proxy support on Linux


aiklimov:
Hi, I'm having trouble getting connections through a web proxy to work on Linux (RHEL 7.1). I'm specifying the proxy by URI:

                web::web_proxy proxy("http://my_proxy_ip_address:my_proxy_port_number");
                web::http::client::http_client_config client_config;
                client_config.set_proxy(proxy);
                web::http::client::http_client client(baseUri, client_config);

I saw that one of the proxy unit tests was marked as Ignore on Linux. Is proxy supported on Linux?
Any help would be appreciated.
-Alex

aiklimov:
This is the unit test I mentioned:
TEST_FIXTURE(uri_address, proxy_with_credentials, "Ignore:Linux", "NYI", "Ignore:Apple", "NYI", "Ignore:Android", "NYI")

Is it the case that only auto-discovery proxies are supported at this time, but not ones with proxy host & port specified?

aiklimov:
Just found this about Linux/non-Windows http_client proxy support:
https://casablanca.codeplex.com/workitem/88

kavyako:
hey aiklimov

Correct, proxy support is Not Yet Implemented on non windows plaforms. Hence the "NIY" tag with the corresponding tests for Linux/Android/Apple.

Thanks
Kavya.

deeringc:
Hi @kavyako,

I'm looking at implementing proxy support for non-Windows platforms. Who would be the best person from your team to have a design discussion with? Or would you rather I just put up a PR once I'm done?

Thanks,
Chris

kavyako:
That's awesome! You could start the discussion here and I will pull in the experts to provide feedback. Don't mind a PR too but going through the design might avoid some roundtripping and save time.

Thanks!
Kavya.

deeringc:
Ok, great! So what I've implemented so far is support for http and https proxies (via ssl tunneling). The design goal I've gone with is to try to continue the patterns that are already used in http_client_asio.cpp and to minimise the impact on existing code/flows so as to reduce the chance of regressions in areas I'm not familiar with. 

In the case of http proxies, the required change is quite small, we basically have to change some of the details in the http request (and add support for basic authentication) but the flow is basically the same.  

[...]

PR #41, opened 2015-12-21
Implementing HTTP and HTTPS client proxy support for non-Windows plat…

deeringc:
As discussed here: https://github.com/Microsoft/cpprestsdk/issues/9

### DIFF
--- a/Release/include/cpprest/details/http_client_impl.h
+++ b/Release/include/cpprest/details/http_client_impl.h
@@ -296,9 +296,10 @@ protected:
     // URI to connect to.
     const http::uri m_uri;
 
-private:
 
-    http_client_config m_client_config;
+	http_client_config m_client_config;
+
+private:
 
     bool m_opened;
 
--- a/Release/include/cpprest/details/http_client_impl.h
+++ b/Release/include/cpprest/details/http_client_impl.h
@@ -296,10 +296,9 @@ protected:
     // URI to connect to.
     const http::uri m_uri;
 
-
-	http_client_config m_client_config;
-
 private:
+
+    http_client_config m_client_config;
 
     bool m_opened;
 
--- a/Release/src/http/client/http_client_asio.cpp
+++ b/Release/src/http/client/http_client_asio.cpp
@@ -74,26 +74,34 @@ class asio_connection
     friend class asio_connection_pool;
     friend class asio_client;
 public:
-    asio_connection(boost::asio::io_service& io_service, bool use_ssl, const std::function<void(boost::asio::ssl::context&)>& ssl_context_callback) :
+    asio_connection(boost::asio::io_service& io_service, bool start_with_ssl, const std::function<void(boost::asio::ssl::context&)>& ssl_context_callback) :
     m_socket(io_service),
     m_pool_timer(io_service),
     m_is_reused(false),
-    m_keep_alive(true)
+    m_keep_alive(true),
+    m_ssl_context_callback(ssl_context_callback)
     {
-        if (use_ssl)
+        if (start_with_ssl)
         {
-            boost::asio::ssl::context ssl_context(boost::asio::ssl::context::sslv23);
-            ssl_context.set_default_verify_paths();
-            ssl_context.set_options(boost::asio::ssl::context::default_workarounds);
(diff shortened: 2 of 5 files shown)


kavyako:
Thank you for the pull request. This will be merged as soon as the project owners get back from vacation.

Release/include/cpprest/details/http_client_impl.h
```diff
@@ -296,9 +296,10 @@ class _http_client_communicator
     // URI to connect to.
     const http::uri m_uri;
 
-private:
 
-    http_client_config m_client_config;
+	http_client_config m_client_config;
```
kavyako: exposing non-`const` access to this data member could be dangerous. The one new use in http_client_asio.cpp can call the member function client_config() for this.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -363,102 +375,331 @@ class asio_context : public request_context, public std::enable_shared_from_this
...
+                {
+                    m_context->report_error("Failed to read HTTP status line from proxy", ec, httpclient_errorcode_context::readheader);
+                }
+            }
         }
+    
+        std::function<void(std::shared_ptr<asio_context>)> m_ssl_tunnel_established;
+        std::shared_ptr<asio_context> m_context;
+    
+        boost::asio::streambuf request_;
```
kavyako: As per our coding conventions, we try to follow m_ prefix for naming class members. It will be good to be consistent.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -477,6 +718,24 @@ class asio_context : public request_context, public std::enable_shared_from_this
...
+    utility::string_t generate_basic_proxy_auth_header()
+    {
+        utility::string_t header;
+        
+        header.append(header_names::proxy_authorization);
+        header.append(": Basic ");
+        
+        auto credential_str = web::details::plaintext_string(new ::utility::string_t(m_http_client->client_config().proxy().credentials().username()));
+        credential_str->append(":");
+        credential_str->append(*m_http_client->client_config().proxy().credentials().decrypt());
```
kavyako: What if credentials are not set.. Should we check for cred::is_set() before accessing the credentials?

Release/src/http/client/http_client_asio.cpp
deeringc: Sure, no problem at all.

Release/src/http/client/http_client_asio.cpp
deeringc: In the places where generate_basic_proxy_auth_header() is called from, the code is checking whether the username is empty or not. I can change that to use is_set() if you think it's better?

Release/include/cpprest/details/http_client_impl.h
deeringc: OK, I'll refactor

Release/src/http/client/http_client_asio.cpp
kavyako: oh I missed the check. Looks like is_set() is also doing the same thing internally, but still it might be better to leverage the API.

kavyako:
Right now we do not have any test cases that exercise the HTTPS proxy code paths. Would it be possible to add a test case? Can we use something like httpbin.org? I think it has HTTPS too, though I am not 100% sure. Another option is to add a manual testcase (with the tag "Ignore", "Manual") and anyone running the testcase can modify the credentials locally.

Release/src/http/client/http_client_asio.cpp
deeringc: Sure, no prob. I agree, it is more readable that way.

deeringc:
Yeah, I was thinking about adding another test case that covers https. I wasn't sure that your internal Microsoft proxy supported this (we don't have access from outside of your network) and didnt want to add a potentially failing test that I couldn't verify myself. I'll do as you suggest and add a manual test. If we come up with some other idea, we can just enable it later.

kavyako:
These changes have been merged to the development branch. Release 2.8.0 (under planning) should have them in the master branch.
Thank you for your contribution.

PR #52, opened 2016-01-12
Updating websocketpp to 6.0-permessage-deflate

AntonBikineev:
Preliminary addressing #33 issue


Danimoth:
Hi Anton,
I did a diff of the code in the PR with the code found here: 
https://github.com/zaphoyd/websocketpp/releases/tag/0.6.0-permessage-deflate

Apart from the change in the second commit of this PR (78bae21), there are the following differences. 

The pull request has this extra file:
Release\libs\websocketpp\websocketpp\error_container.hpp

The official (pre-)release has these extra files:
websocketpp-0.6.0-permessage-deflate\test\http\a.out
websocketpp-0.6.0-permessage-deflate\test\http\perf.out
websocketpp-0.6.0-permessage-deflate\test\http\test.out
websocketpp-0.6.0-permessage-deflate\cmake\CMakeHelpers.cmake

And the only different file is:
websocketpp-0.6.0-permessage-deflate\.travis.yml

where the official one installs libboost-chrono1.48-dev in addition to other boost1.48 stuff, but the one in the pull request doesn’t.  

Can you briefly explain the changes you made (other than the 78bae21 which is explained in the commit message)? Thanks!

AntonBikineev:
@Danimoth 
Hi Alexander,

First, thanks for response.

Release\libs\websocketpp\websocketpp\error_container.hpp
This file seems to be leftover from previous version. My bad that I didn't check for removed files in new version.

All *.out files were ignored by my .gitignore. CMakeHelpers.cmake was ignored too, but not sure whether it is really used. Will add to the PR anyway.

The same for .travis.

Thanks for all points.

ras0219-msft:
We noticed that websocketpp 0.7.0 released, so we're going to try to update to that one instead.

https://github.com/zaphoyd/websocketpp/releases

ras0219-msft:
We now support using an external websocketpp which can be independently upgraded (searched for via `find_package(WEBSOCKETPP CONFIG)`).

PR #82, opened 2016-03-03
Fixed ambiguous call to begin when using with boost library.

dogukancagatay:
When using the cpprestsdk with Boost library, compiler was complaining about ambiguous call in containerstream.h on Windows.

> containerstream.h boost::range_adl_barrier::begin ambiguous call to overloaded function


ras0219-msft:
Thank you for this fix! I have merged it into the development branch.

PR #86, opened 2016-03-09
Fix obtaining raw string_t pointer from temporary

schellap:


ras0219-msft:
Thank you for this fix! It has been merged into the development branch.

PR #90, opened 2016-03-11
make soname only have major version number

hanzhumsft:
Suppose cpprestsdk follows semver, the soname doesn't need to have minor version number.


ras0219-msft:
Unfortunately, we do not follow semver for binary ABI. We only follow semver for source-code APIs. We therefore must include the entire version number in the SONAME.

PR #102, opened 2016-03-23
Added websocket_client_config option for ssl verify mode

hyotak-yun:
disable certificate validation for websocket


ras0219-msft:
This looks good. However, we need to add tests to verify that the functionality works as intended.

kavyako:
This is merged to development branch. I will add a test case to verify the scenario.

Issue #61, opened 2016-01-25
Build failures on Ubuntu 14.04 with clang 3.4


ms705:
On a fully up-to-date Ubuntu 14.04 LTS installation, I get the following error when building cpprestsdk (specifically, `test_runner`) with clang instead of GCC:

```shell
$ cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang
[...]
-- The C compiler identification is Clang 3.4.0
-- The CXX compiler identification is Clang 3.4.0
-- Check for working C compiler: /usr/bin/clang
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++
-- Check for working CXX compiler: /usr/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Boost version: 1.55.0
-- Found the following Boost libraries:
--   random
--   chrono
--   system
--   thread
--   regex
--   filesystem
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libssl.so;/usr/lib/x86_64-linux-gnu/libcrypto.so (found suitable version "1.0.1f", minimum required is "1.0.0")
-- Setting clang options
-- websocketpp not found, using the embedded version
-- Added test library httpclient_test
-- Added test library httplistener_test
-- Added test library json_test
-- Added test library pplx_test
-- Added test library streams_test
-- Added test library uri_test
-- Added test library utils_test
-- Added test library websocketclient_test
-- Configuring done
-- Generating done
-- Build files have been written to: /src/cpprestsdk-git/Release/build.release
$
$ make
[  0%] Building CXX object src/CMakeFiles/cpprest.dir/http/client/http_client_msg.cpp.o
[...]
Linking CXX shared library ../Binaries/libcpprest.so
[ 19%] Built target cpprest
[...]
[...]

PR #114, opened 2016-03-30
Removed redundant std::move() that was causing errors on Xcode 7.3 gcc

leighsmith:
According to [this link](https://www.ibm.com/developerworks/community/blogs/5894415f-be62-4bc0-81c5-3956e82276f3/entry/RVO_V_S_std_move?lang=en) returning std::move() isn't the right way to do things, and breaks on Xcode 7.3 compiler.

### DIFF
--- a/Release/src/http/common/http_helpers.cpp
+++ b/Release/src/http/common/http_helpers.cpp
@@ -272,7 +272,7 @@ utf16string convert_utf16_to_utf16(utf16string src)
     {
     case little_endian:
         src.erase(0, 1);
-        return std::move(src);
+        return src;
     case big_endian:
         return convert_utf16be_to_utf16le(std::move(src), true);
     case unknown:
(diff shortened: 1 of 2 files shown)


ras0219-msft:
Thank you very much for the notice and fix. :+1:

eagleflo:
I encountered this while developing against the master branch. Someone also brought it up in https://github.com/Microsoft/cpprestsdk/issues/61 recently.

IMO you could release 2.8.1 with this fix in it.

mmcknett:
FWIW, this is the first thing you encounter on OSX when trying to get started with the library (while following the ["How to build" instructions](https://github.com/Microsoft/cpprestsdk/wiki/How-to-build-for-Mac-OS-X)).  Releasing this fix might be worth the good first impression.  :)

guysherman:
I have experienced this issue as well. I agree that this should be rolled out as a 2.8.1 release straight away.

Issue #101, opened 2016-03-23
utility::conversions::latin1_to_utf16 - Is there a bug for characters above 0x80?


pp555:
I've tried to use `utility::conversions::latin1_to_utf16` to convert latin1 string to UTF-16. The string contained 'í' character (0xed). After converting it should become 0x00ed but it seems that converting method replaced it with 0xffed.

I think this may be result of casting signed `char` to `utf16char` and `dest[i] = utf16char(s[i]);` in asyncrt_utils.cpp should be replaced by `dest[i] = utf16char(static_cast<unsigned char>(s[i]));` to fix it. Can someone check if I'm right?

ras0219-msft:
This could very well be an issue. Please consider making a PR with the fix and a test case; it would be gladly accepted.

PR #116, opened 2016-03-31
Fixing latin1 to UTF-16 convertion

pp555:
As I described in #101, I found bug in `utility::conversions::latin1_to_utf16`, when trying to convert string containing 'í' character (0xed). Signed character is being casted and it leads to result containing 0xff as first byte instead of 0x00 (example for input char 0xed: 0xffed instead of 0x00ed which is correct UTF-16 encoded 'í' character).
This bug affects strings containing any characters above 0x80 (negative numbers in signed `char`).


ras0219-msft:
Thanks for this PR!

PR #117, opened 2016-03-31
Fix header reading on linux listener using HTTPS.

slepasteur:
The first asynchronous read of the stream was stopping at the first CRLF (after the HTTP version) instead of reading until CRLFCRLF (the end of the header).

This bug was not always noticeable because, as stated in the boost documentation:
"After a successful async_read_until operation, the streambuf may contain additional data beyond the delimiter. An application will typically leave that data in the streambuf for a subsequent async_read_until operation to examine."
Therefore the bug appeared only when trying to read the body of the request and the streambuf filled by async_read_until did not contain the complete header.

### DIFF
--- a/Release/src/http/listener/http_server_asio.cpp
+++ b/Release/src/http/listener/http_server_asio.cpp
@@ -58,7 +58,7 @@ namespace details
 // This is used as part of the async_read_until call below; see the
 // following for more details:
 // http://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/reference/async_read_until/overload4.html
-struct crlf_nonascii_searcher_t
+struct crlfcrlf_nonascii_searcher_t
 {
     enum class State
     {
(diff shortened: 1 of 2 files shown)


ras0219-msft:
Is there some way to test this? Our current tests clearly don't cover this case.

slepasteur:
You could make a request on an HTTPS listener with a lot of header and check that all the header are read correctly.

I tried to modify the create_https_listener_get test to include this but failed (the test crashed when I added a lot of header to the request).

ras0219-msft:
Hmm. Perhaps a different bug? What did you try doing?

slepasteur:
Ok I managed to create the test. I don't know what I was doing wrong first. 

I added all the headers found on [wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields) to the request. This generates enough content to be sure* that the headers will not be read by chance. I then checked that the body and all the headers were read correctly.

* the size that might be present in the streambuf is not defined so it might not be enough in some case

ras0219-msft:
Thank you very much for adding the test case, and an overall great PR. Sorry for the long delay on merging here!

PR #123, opened 2016-04-11
Fix a memory leak issue

EvanCui:
This change fixed a memory leak in the implementation of connection pool.
The original implementation uses a ref count for managing the lifetime of each connection, but the connection object is definitely lost reported by valgrind.
Investigation showing that the m_refs field is added by do_response twice, so in the end of finish_request_response method, it equals 1.
This change fixed the issue by just delete the instance at the end of finish_request_response method.

### DIFF
--- a/Release/src/http/listener/http_server_asio.cpp
+++ b/Release/src/http/listener/http_server_asio.cpp
@@ -602,13 +602,10 @@ void connection::dispatch_request_to_listener()
             m_request._reply_if_not_already(status_codes::InternalError);
         }
     }
-    
-    if (--m_refs == 0) delete this;
 }
 
 void connection::do_response(bool bad_request)
 {
-    ++m_refs;
     m_request.get_response().then([=](pplx::task<http::http_response> r_task)
     {
         http::http_response response;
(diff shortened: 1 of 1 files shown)


ras0219-msft:
First, thanks for finding this issue.

With this change, the refcount is never incremented: it's initialized to 1 in the constructor and then decremented inside finish_request_response. Basically, it would act like there isn't any refcounting at all! 

While this is certain to delete the object and prevent it leaking, I am very concerned about the possibility of use-after-free. It's hard to believe this is correct given the level of concurrency here.

For example: `do_response(false)` is called in the middle of `dispatch_request_to_listener()`, which then goes on to use member variables in the following lines. Inside `do_response()`, an asynchronous task is launched that will eventually call `finish_request_response()`. Because this is launched on another thread, that task could be scheduled to complete before the remaining tasks in `dispatch_request_to_listener()` completes! This would therefore result in use-after-free.

Obviously as it stands there is a memory leak as valgrind found. However, I'm certain that the fix you've proposed would add the potential for a use-after-free. A better solution would probably be to avoid using this error-prone manual refcounting entirely and switch to using std::shared_ptr<> instead.

EvanCui:
Agreed with the std::shared_ptr<> solution.

For the current fix, we have tested in a stress testing, which is stable. But eventually, I will move to shared_ptr.

ras0219-msft:
I've reworked the internal asio server refcounting to remove this leak in a way that should more closely guarantee correctness in 38d916a6.

PR #131, opened 2016-04-17
Update to include access control allow origin

jsonng:
added DAT(access_control_allow_origin,  "Access-Control-Allow-Origin") as a list of header items.


ras0219-msft:
Thanks for the PR even with this being such a small change :).

PR #135, opened 2016-04-22
Add certificate pinning capabilities to WinHttp

glukacsy:
This is a preliminary pull request to see if you guys are happy to add this to Casablanca. Essentially, due to security requirements, we are using certificate pinning and needed to expose this functionality in Casablanca. 

On a high level all we are doing here is letting the consumer register a callback which will be called back for each certificate retrieved during the TLS negotiation. The consumer then compare these against a hardcoded list of valid public keys and make a yes/no decision. If all certificates are rejected, the connection is cancelled. 

The reason we need to walk the chain (and which makes the code complex) is that in most cases we don't want to pin to the leaf cerrtificate, but rather to an intermediate one (because the leaf certificates change too often). 

Let me know if on a high level this is in line with what you have in mind regarding design. In the meantim I will work on doing the same for OS X.

Regards,
Gergely

### DIFF
--- a/Release/src/http/client/http_client_winhttp.cpp
+++ b/Release/src/http/client/http_client_winhttp.cpp
@@ -29,6 +29,9 @@
 #include "cpprest/http_headers.h"
 #include "http_client_impl.h"
 
+#include <wincrypt.h> // for certificate pinning logic
+#include <iomanip> // for certificate pinning logic
+
 namespace web
 {
 namespace http
@@ -462,7 +465,7 @@ protected:
         if(WINHTTP_INVALID_STATUS_CALLBACK == WinHttpSetStatusCallback(
             m_hSession,
             &winhttp_client::completion_callback,
-            WINHTTP_CALLBACK_FLAG_ALL_COMPLETIONS | WINHTTP_CALLBACK_FLAG_HANDLES,
+			WINHTTP_CALLBACK_FLAG_ALL_COMPLETIONS | WINHTTP_CALLBACK_FLAG_HANDLES |	WINHTTP_CALLBACK_FLAG_SEND_REQUEST,
             0))
         {
             return report_failure(_XPLATSTR("Error registering callback"));
@@ -898,6 +901,19 @@ private:
         }
     }
 
+	static utility::string_t getCertificateString(BYTE* publicKey, DWORD size)
+	{
+		// obtain public key of certificate as string
+		std::stringstream stream;
+		stream << std::hex;
+		for (size_t i = 0; i < size; ++i)
+		{
+			stream << std::setw(2) << std::setfill('0') << (int)((unsigned char*)publicKey)[i];
+		}
+		
+		return utility::conversions::to_string_t(stream.str());
+	}
+
     // Returns true if we handle successfully and resending the request
     // or false if we fail to handle.
     static bool handle_authentication_failure(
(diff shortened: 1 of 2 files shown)


Release/src/http/client/http_client_winhttp.cpp
```diff
@@ -1053,6 +1069,112 @@ class winhttp_client : public _http_client_communicator
                     p_request_context->report_error(errorCode, build_error_msg(error_result));
                     break;
                 }
+			case WINHTTP_CALLBACK_STATUS_SENDING_REQUEST:
```
glukacsy: This events comes twice, according to the documentation (https://msdn.microsoft.com/en-us/library/windows/desktop/aa384115%28v=vs.85%29.aspx): "Activates upon beginning and completing the sending of a request header with WinHttpSendRequest." 

This means we are doing pinning check twice for any HTTP requests - i'll see if there is any way to differentiate between the two occurences of the same callback.

ras0219-msft:
Certificate pinning is certainly an important feature for all HTTP clients and it'd be great to make this easier for consumers.

I see two general approaches to these sorts of problems:
1) Enable users to write platform-specific code. Expose the underlying callback interfaces so clients can access the full richness of the underlying APIs. This approach minimizes implementation burden on our side and maximizes flexibility for the user, but doesn't make it easy for the user to do the right thing.
2) Provide a completely general interface that maximizes ease of use by targeting common scenarios. This requires a separate interface for every feature, but is better for users.

I feel like your API so far straddles between the two. While it's certainly true that you aren't passing native handles through the callback, it looks like the API is still lower level than what most pinning programs would actually require. I'd imagine that most programs simply require a way to register a single certificate hash; in rare cases multiple.

Additionally, I'm not sure why the hostname is _directly_ relevant here (though I could be wrong). It seems like regardless of pinning you still want the normal "ensure hostname matches CN" check to be performed. If you don't know the destination URI ahead of time, what would be wrong with just providing the hash of every valid destination URI's cert? I assume that if you can't provide them ahead of time, you wouldn't be able to correctly check them in a callback either.

Since you're designing this feature due to your own requirements, could you provide a bit more detail on how exactly you need the certificates pinned and why the callback interface works better than registering a set of hashes? I don't mean to discourage; just improving my understanding of the pinning use cases.

glukacsy:
Thanks for coming back so quickly. I agree with your categorisation and based on what you are suggesting I am happy to rework the flows so the interface takes a list of (public key) signatures and casablanca takes care of the rest internally without any callbacks required. 

Regarding your question about the hostname: you are right, this might be specific to our application, but I think we can actually generalise it. The reason I need the actual host is because we don't want to pin for every host, there are exceptions. And specifically, in case of a redirect, it is possible that the redirected URL falls into one of the exceptions where we only want to do the traditional certificate checking (which casablanca already does by default), but not pinning. 

Typical example: client connects to a known address where we normally want to pin, but one of the endpoints on that micro service actually redirects to rackspace (for example). We cannot and don't want to pin the cert for rackspace and a such currently we just report success in the pinning callback if the URL goes to rackspace.

I guess the way we could generalise is to register - for a given http config - a set of public keys as well as a list of exceptions and then let cpprest do the rest? How does that sound?

Ah, one more thing:   instead of p_request_context->cleanup(); I tried to raise a specific error, but it turns out we run into a double delete there: the request context will be freed and even though the code deregisters the winhttp callback, we are still getting a WINHTTP_CALLBACK_STATUS_REQUEST_ERROR which then throws up due to accessing a dead object. I am wondering if there is a generic issue in http_client_winhttp.cpp for any other case when we are throwing an exception/error within completion_callback.

glukacsy:
(the dead object is the context argument in completion_callback)

ras0219-msft:
After reading more into pinning, I see that having the hostname is really quite essential. Thank you for your example; it makes things much clearer.

Looking at some examples from other languages given on https://www.owasp.org/index.php/Pinning_Cheat_Sheet, I think that the APIs used in http://datatheorem.github.io/TrustKit/ seem very easy to use and rather complete as well. Specifically,

```swift
let trustKitConfig = [
   kTSKPinnedDomains: [
       "yahoo.com": [
           kTSKPublicKeyAlgorithms: [kTSKAlgorithmRsa2048],
           kTSKPublicKeyHashes: [
               "JbQbUG5JMJUoI6brnx0x3vZF6jilxsapbXGVfjhN8Fg=",
               "WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18="
             ],]]]
  
TrustKit.initializeWithConfiguration(config)
```
I think having a multimap-like interface in the `http_client_config` object that lets you register a set of hostname -> "requirements" relations (no exceptions required) would be a solid API. If we want to support additional verification options in the future such as specifying the encryption type or other parameters, it should be straightforward to add fields to the "requirements" struct beyond just the hash.

Something the above wouldn't support is the ability to have a hostname "whitelist" (so all other hosts get blocked). However, I think that kind of feature may be better served by a separate API. In your scenario, do you have any cases where "everything not pinned follows the normal rules" is not acceptable?

Regarding a potential bug with `p_request_context->cleanup()` -- it does look like there could easily be lifetime issues here. We should probably be using a weak_ptr instead of a raw pointer as the callback context to allow the object to be freed while callbacks are outstanding.

glukacsy:
The "everything not pinned follows the normal rules" works for us in general.  Regarding the above API. I think TrustKit has that multi map as a global configuration for all connections (it applies to all NSURLConnection for example). I am not sure doing that on an http_client_config level is the best (as an http_client_config usually belongs to a specific http_client instance, proxy settings etc for the given address specific to the http_client base url).  

If we take redirect out of the picture, I think the cleanest API would be to simply add a list of public keys to an http_client_config object. If that list is empty no pinning is applied, if it's not empty, pinning is  applied. 

The complications come in when there is redirect involved. So there we need to expose some way of controlling what the behaviour should be when there is a redirect (or multiple redirects).

The other option is to not to use http_client_config, but rather try to set it globally (similar to TrustKit), but then we might need to provide a way to turn it off on a per http_client basis (if that's needed) and in general it's not clear to me if there is a concept of such a global store in casablanca (or whether it's even a good thing to have).

So coming back in full circle, I think the original callback design has the advantage that it bypasses some of the issues above and gives more control to the consumers (in fact lets the consumer easily centralise the pinning logic vs spread it out when creating different http_client_config instances with different set of multi map registrations etc).

ras0219-msft:
While my knee-jerk reaction is against a global store, as I think about the particulars it does make a certain degree of sense. For a given host, if you _ever_ want to pin a set of certs, it seems like you would _always_ want to pin those certs across the entire process. Or, put another way, is there ever a time where you would want to simultaneously create one http_client that pins but another (to the same host) that doesn't? I can't think of any.

Setting that side, however: what do you think about adding a lower-level callback instead? In this case it would be hooking WINHTTP_CALLBACK_STATUS_SENDING_REQUEST directly.

We could then add the bulk of your implementation as a kind of callback wrapper that converts the low-level callback into a higher-level `string, string` callback like you've proposed above. This additionally allows us to add the callback wrapper into the experimental namespace, which will make the burden of testing easier.

I didn't bring this up before, but in order to add a public, stable security-related API I would really want quite hefty test coverage over all the possibilities (especially the edge cases like missing leaf certificates and invalid intermediates). Adding the bulk of the code into an experimental namespace would be much easier from a stability point of view.

What do you think about that?

glukacsy:
Sounds good to me, is there an example for such a callback wrapper? 

Just to clarify, where would the (string,string) callback be exposed if not on the http_config object? 

In other words, as an end user, how would the usage of this feature in the experimental namespace look like?

glukacsy:
Poke.... In the meantime we have wired up the pinning on non windows platforms using http_client_asio and also added pinning to websockets. Do you want a PR for those - the reason I am asking is that they rely on the current design and I am not sure if you want to go ahead with that (based on the discussion above).

ras0219-msft:
I would imagine it as some struct + helper function in the experimental namespace with an appropriate `operator()`. Something like

```
namespace experimental
{
template<class F>
struct basic_pinning_helper_t
{
    basic_pinning_helper(F f) : inner_func(f) {}
    bool operator()(HINTERNET hRequestHandle) { /* call the user's inner_func in here */ }
    F inner_func;
};
template<class F>
auto basic_pinning_helper(F f) { return basic_pinning_helper_t<F>(f); }
}
```

The user would then do something like
```
client.set_send_request_callback(
    experimental::basic_pinning_helper(
        [](string& host, string& digest)
        {
            /* Your logic here */
        }));
```

matra774:
Hi, any plans on merging this? This functionallity is quite usefull...

Release/src/http/client/http_client_winhttp.cpp
coin8086: No, it's just once. The document you mentioned is for WINHTTP_CALLBACK_STATUS_SEND_REQUEST, which is defined as a combination of two events:
```c++
#define WINHTTP_CALLBACK_FLAG_SEND_REQUEST              (WINHTTP_CALLBACK_STATUS_SENDING_REQUEST | WINHTTP_CALLBACK_STATUS_REQUEST_SENT)
```

Release/src/http/client/http_client_winhttp.cpp
```diff
@@ -462,7 +465,7 @@ class winhttp_client : public _http_client_communicator
         if(WINHTTP_INVALID_STATUS_CALLBACK == WinHttpSetStatusCallback(
             m_hSession,
             &winhttp_client::completion_callback,
-            WINHTTP_CALLBACK_FLAG_ALL_COMPLETIONS | WINHTTP_CALLBACK_FLAG_HANDLES,
+			WINHTTP_CALLBACK_FLAG_ALL_COMPLETIONS | WINHTTP_CALLBACK_FLAG_HANDLES |	WINHTTP_CALLBACK_FLAG_SEND_REQUEST,
```
coin8086, line 468: I guess it should be WINHTTP_CALLBACK_STATUS_SENDING_REQUEST instead of WINHTTP_CALLBACK_FLAG_SEND_REQUEST.

PR #136, opened 2016-04-24
Revert "Removing no longer needed files for building libiconv for Android"

### LINKED ISSUE #85: Build_android/configure.sh references non-existant libiconv.patch
It appears that either the libiconv.patch didn't get checked in or is no longer necessary but is still referenced.

denizt:
This reverts commit 0dd86b11535d8db64c5a7409b1622d71bfbb21f6.

It also adds the missing boost atomic and bumps the clang to use the most recent version. With these changes, the compilation for android under Linux now works as expected.

Fixes #85

### DIFF
--- a/Build_android/configure.sh
+++ b/Build_android/configure.sh
@@ -187,7 +187,7 @@ then
 	    git apply "$DIR/boost-for-android.patch"
 	    touch cpprestsdk.patched.stamp
 	fi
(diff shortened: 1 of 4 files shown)


denizt:
On a second look, it seems it is no longer needed however there were traces left breaking build. We should probably wipe out all of its traces (i.e. --skip-iconv)

PR #137, opened 2016-04-25
Fix android build script for linux

### LINKED ISSUE #85: Build_android/configure.sh references non-existant libiconv.patch

denizt:
The build no longer uses libiconv but there were traces left. Also boost patch had a typo for x86. With these changes, the build script should work as expected.

Fixes #85

### DIFF
--- a/Build_android/boost-for-android-x86.patch
+++ b/Build_android/boost-for-android-x86.patch
@@ -13,29 +13,17 @@ index 40453f7..7ffc050 100755
 +		CXXPATH=$AndroidNDKRoot/toolchains/${TOOLCHAIN}/prebuilt/${PlatformOS}-x86_64/bin/clang++
 +		TOOLSET=clang-androidR8e
 +                ;;
-+	"10e-rc4 (64-bit)")
++	"10e-rc4 (64-bit)"|"10e (64-bit)")
 +		TOOLCHAIN=llvm-3.6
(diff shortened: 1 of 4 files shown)


ras0219-msft:
Thanks for the incredibly thorough PR here; I'm pretty sure you covered all the references to libiconv. Nice work!

I am curious about one change: you uncommented the `<architecture>i686` line. While it certainly wasn't doing anything useful before, I'm curious what particularly motivated the change? Did you get build errors without it?

denizt:
Thanks! The removal of comment out is a byproduct. That being said, I do have problems linking this thing, still. I wholeheartedly recommend bumping the boost to a more recent version and replacing 'Boost-for-Android' with vanilla boost, we are using 1.59 without problems on android at work.

Also there seems to be a small issue with the edited patch. I will push a follow up tonight.

ras0219-msft:
Please feel free to make a separate pull request that replaces Boost-for-Android with vanilla boost. Reducing dependencies is always a good thing.

denizt:
With these changes both platforms should build (just tested that it builds). I will make a separate PR for boost which should remove the need for most of this PR.

ras0219-msft:
Thank you very much for this PR, removing unused code is really important 😊.

cipriancaba:
Any reason for this not to be merged into master?

denizt:
@cipriancaba This has already been merged to development branch @ https://github.com/Microsoft/cpprestsdk/tree/development

cipriancaba:
Makes sense, thank you

Issue #67, opened 2016-02-03
Annoying CRT memory leak reports with 2.7.0 build from NuGet


vadz:
After upgrading a simple project using the library from VS2013 with C++ REST SDK 2.1.0 to VS2015 with 2.7.0 I started getting
```
Detected memory leaks!
Dumping objects ->
{345} normal block at 0x014A6970, 12 bytes long.
 Data: <       Yp  Y> 01 00 00 00 D0 9D E8 59 70 B4 F4 59 
{344} normal block at 0x014A5C28, 16 bytes long.
 Data: <piJ             > 70 69 4A 01 00 00 00 00 00 00 00 00 00 00 00 00 
{343} normal block at 0x014A5BA8, 20 bytes long.
 Data: <    (\J         > 01 00 00 00 28 5C 4A 01 01 00 00 00 04 00 00 00 
Object dump complete.
```
on application exit. The allocation stack for the allocation 343 looks like this:
```
>	ucrtbased.dll!malloc(unsigned int size=0x00000014) Line 18	C++
 	cpprest140d_2_7.dll!default_malloc_ex(unsigned int num=0x00000014, const char * file=0x5f226ba4, int line=0x000000a2) Line 79	C
 	cpprest140d_2_7.dll!CRYPTO_malloc(int num=0x00000014, const char * file=0x5f226ba4, int line=0x000000a2) Line 342	C
 	cpprest140d_2_7.dll!sk_new(int(*)(const void *, const void *) c=0x5f132460) Line 162	C
 	cpprest140d_2_7.dll!load_builtin_compressions() Line 483	C
 	cpprest140d_2_7.dll!SSL_COMP_get_compression_methods() Line 1949	C
 	cpprest140d_2_7.dll!SSL_library_init() Line 153	C
 	cpprest140d_2_7.dll!boost::asio::ssl::detail::openssl_init_base::do_init::do_init() Line 40	C++
 	cpprest140d_2_7.dll!boost::asio::ssl::detail::openssl_init_base::instance() Line 131	C++
 	cpprest140d_2_7.dll!boost::asio::ssl::detail::openssl_init<1>::openssl_init<1>() Line 60	C++
 	cpprest140d_2_7.dll!`dynamic initializer for 'boost::asio::ssl::detail::openssl_init<1>::instance_''() Line 90	C++
 	ucrtbased.dll!_initterm(void(*)() * first=0x5f206000, void(*)() * last=0x5f206684) Line 22	C++
```
[...]

PR #143, opened 2016-05-06
Work around SSL compression methods memory leak in ASIO

vadz:
Call `SSL_COMP_free_compression_methods()` because ASIO itself doesn't, even
though it should, because it calls `SSL_library_init()` which allocates memory
for the compression methods.

Closes https://github.com/Microsoft/cpprestsdk/issues/67

---
This is ugly but seems like the only way to fix the annoying leak reports other than fixing the [bug in ASIO](https://svn.boost.org/trac/boost/ticket/10795) itself, but considering that Casablanca still uses Boost 1.55, I'm not sure how soon will it get this fix even if it does get fixed upstream.


ras0219-msft:
Thank you for this PR. I think this is probably a good fix for the time being.

Two changes though:
1) Please make the PR against the development branch
2) We only use ASIO on windows for websockets, so this should probably go into `Release/src/websockets/client/ws_client_wspp.cpp`. The only platform that doesn't pull in that file is WinRT, which doesn't use ASIO.

vadz:
@ras0219-msft Thanks for your quick review! I've update the PR and retested that it still fixes problem.

playbass:
I tried a build of the current development branch and it didn't fix the memory leaks. Only by defining CPPREST_EXCLUDE_WEBSOCKETS was I able to stop them. Is there something I'm missing regarding the fix mentioned above regarding the ASIO memory leaks?

PR #146, opened 2016-05-09
Fix ambiguous reference to ‘credentials’

dkelson:
Allows compilation on Linux with GCC v5.3


ras0219-msft:
Already fixed in development.

PR #149, opened 2016-05-10
Some perf improvements for uri related code

hanzhumsft:
- added overloaded function of print_string and scan_string to have string type specialized.
- added overloaded uri constructor to allow building from uri_components directly, which can save one pass of uri parsing.
- reduced the use of std::locale::classic(), which cost much CPU in thread synchronization, in places where only US-ASCII is involved.
- replace stringstream with string::append, as stringstream is notably slow in concatenating just plain strings because it has a big overhead of formatting which is designed for concatenating heterogeneous objects.

### DIFF
--- a/Release/include/cpprest/base_uri.h
+++ b/Release/include/cpprest/base_uri.h
@@ -227,7 +227,7 @@ namespace web {
         /// Creates a URI from the given URI components.
         /// </summary>
         /// <param name="components">A URI components object to create the URI instance.</param>
-        _ASYNCRTIMP uri(const details::uri_components &components) : m_components(components) { m_uri = m_components.join(); }
+        _ASYNCRTIMP uri(const details::uri_components &components);
 
         /// <summary>
         /// Creates a URI from the given encoded string. This will throw an exception if the string
--- a/Release/include/cpprest/http_headers.h
+++ b/Release/include/cpprest/http_headers.h
@@ -159,7 +159,7 @@ public:
     {
         if (has(name))
         {
-            m_headers[name] =  m_headers[name].append(_XPLATSTR(", ")).append(utility::conversions::print_string(value));
+            m_headers[name].append(_XPLATSTR(", ")).append(utility::conversions::print_string(value));
         }
         else
         {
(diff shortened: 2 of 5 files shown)


Release/include/cpprest/base_uri.h
```diff
@@ -224,6 +224,12 @@ namespace web {
         uri() { m_uri = _XPLATSTR("/");};
 
         /// <summary>
+        /// Creates a URI from the given URI components.
+        /// </summary>
+        /// <param name="components">A URI components object to create the URI instance.</param>
+        _ASYNCRTIMP uri(const details::uri_components &components) : m_components(components) { m_uri = m_components.join(); }
```
ras0219-msft: If the function is defined inline, it shouldn't be marked _ASYNCRTIMP. This can be fixed when merging.

Release/include/cpprest/http_headers.h
```diff
@@ -159,7 +159,7 @@ class http_headers
     {
         if (has(name))
         {
-            m_headers[name] =  m_headers[name].append(_XPLATSTR(", ") + utility::conversions::print_string(value));
+            m_headers[name] =  m_headers[name].append(_XPLATSTR(", ")).append(utility::conversions::print_string(value));
```
ras0219-msft: Since `string::append()` modifies the string, we no longer need an assignment back into the map here. This can be fixed when merging.

Release/include/cpprest/base_uri.h
hanzhumsft: Yes. You are right.

Release/include/cpprest/http_headers.h
hanzhumsft: Right, we can just remove the assignment.

ras0219-msft:
This change regresses two URI tests.
```
**** uri_builder_tests:to_string_invalid_uri FAILED ****
**** uri_builder_tests:append_query_locale FAILED ****
```
Both of them repro 100% on Windows Desktop (VS 2015 x86 Debug).

They will need to get fixed before I can merge this.

hanzhumsft:
Finished running all 824 tests.

Only **** proxy_tests:proxy_with_credentials FAILED ****
But this test is not related to my change.

PR #156, opened 2016-05-16
add host based connection pool map

hanzhumsft:
This change is for the http_client implementation on Linux.
- Add a map for host to connection pool mapping.
- Asio connection pool is not destroyed when http_client object is destroyed.
- Multiple http_client objects can use the same connection pool if the base_uri is the same.

The change can make the following code running efficiently on Linux.
```c++
make_request(/*a list of requests*/ l)
{
    request r = l.head;
    http_client client(r.request_uri().authority());
    client.request(r).then( ...; make_request(l.tail); );
}
```
Originally, the following code will create a connection, close it, create another connection, and so on ... 
Now the connections can be reused if the authority is same.

On Windows, connection is always reused (based on host) and managed by WinHTTP. This change makes it easier for cross-platform projects to use the same piece of code.

### DIFF
--- a/Release/include/cpprest/details/http_server_asio.h
+++ b/Release/include/cpprest/details/http_server_asio.h
@@ -87,7 +87,10 @@ public:
         if (is_https)
         {
             m_ssl_context = utility::details::make_unique<boost::asio::ssl::context>(boost::asio::ssl::context::sslv23);
-            ssl_context_callback(*m_ssl_context);
+            if (ssl_context_callback)
+            {
+                ssl_context_callback(*m_ssl_context);
+            }
             m_ssl_stream = utility::details::make_unique<boost::asio::ssl::stream<boost::asio::ip::tcp::socket&>>(*m_socket, *m_ssl_context);
 
             m_ssl_stream->async_handshake(boost::asio::ssl::stream_base::server, [this](const boost::system::error_code&) { this->start_request_response(); });
--- a/Release/include/cpprest/http_client.h
+++ b/Release/include/cpprest/http_client.h
@@ -100,7 +100,6 @@ public:
 #endif
         , m_set_user_nativehandle_options([](native_handle)->void{})
 #if !defined(_WIN32) && !defined(__cplusplus_winrt)
-        , m_ssl_context_callback([](boost::asio::ssl::context&)->void{})
         , m_tlsext_sni_enabled(true)
 #endif
 #if defined(_WIN32) && !defined(__cplusplus_winrt)
--- a/Release/include/pplx/threadpool.h
+++ b/Release/include/pplx/threadpool.h
@@ -42,10 +42,6 @@
 #include "pplx/pplx.h"
 #endif
 
-namespace web { namespace http { namespace client { namespace details {
-    class asio_connection_pool;
-}}}}
-
 namespace crossplat {
 
 #if (defined(ANDROID) || defined(__ANDROID__))
@@ -100,31 +96,6 @@ public:
     {
         return m_service;
     }
-
-    template<typename PoolGenerator>
-    std::shared_ptr<web::http::client::details::asio_connection_pool> obtain_connection_pool(const std::string &key, PoolGenerator pool_generator)
-    {
-        std::lock_guard<std::mutex> lg(m_connection_pool_map_mutex);
-
-        auto &pool = m_connection_pool_map[key];
-        if (!pool)
-        {
-            pool = pool_generator();
-        }
-
-        return pool;
-    }
-
-    template<typename PoolReleaseHandler>
-    void release_connection_pool(const std::string &key, PoolReleaseHandler handler)
-    {
-        std::lock_guard<std::mutex> lg(m_connection_pool_map_mutex);
-
-        auto pool = m_connection_pool_map[key];
-        handler(pool);
-    }
-
-    void free_connection_pool(const boost::system::error_code &ec, const std::string &key);
 
 private:
     struct _cancel_thread { };
@@ -185,9 +156,6 @@ private:
     std::vector<pthread_t> m_threads;
     boost::asio::io_service m_service;
     boost::asio::io_service::work m_work;
-
-    std::mutex m_connection_pool_map_mutex;
-    std::map<std::string, std::shared_ptr<web::http::client::details::asio_connection_pool>> m_connection_pool_map;
 };
 
 }
--- a/Release/include/pplx/threadpool.h
+++ b/Release/include/pplx/threadpool.h
@@ -101,18 +101,30 @@ public:
         return m_service;
     }
 
-    std::shared_ptr<web::http::client::details::asio_connection_pool> obtain_connection_pool(const std::string &base_uri, std::function<std::shared_ptr<web::http::client::details::asio_connection_pool>()> connection_pool_generator)
+    template<typename PoolGenerator>
+    std::shared_ptr<web::http::client::details::asio_connection_pool> obtain_connection_pool(const std::string &key, PoolGenerator pool_generator)
     {
         std::lock_guard<std::mutex> lg(m_connection_pool_map_mutex);
 
-        auto &pool = m_connection_pool_map[base_uri];
+        auto &pool = m_connection_pool_map[key];
         if (!pool)
         {
-            pool = connection_pool_generator();
+            pool = pool_generator();
         }
 
         return pool;
     }
+
+    template<typename PoolReleaseHandler>
+    void release_connection_pool(const std::string &key, PoolReleaseHandler handler)
+    {
+        std::lock_guard<std::mutex> lg(m_connection_pool_map_mutex);
+
+        auto pool = m_connection_pool_map[key];
+        handler(pool);
+    }
+
+    void free_connection_pool(const boost::system::error_code &ec, const std::string &key);
 
 private:
     struct _cancel_thread { };
--- a/Release/include/pplx/threadpool.h
+++ b/Release/include/pplx/threadpool.h
@@ -42,6 +42,10 @@
 #include "pplx/pplx.h"
 #endif
 
+namespace web { namespace http { namespace client { namespace details {
+    class asio_connection_pool;
+}}}}
+
 namespace crossplat {
 
 #if (defined(ANDROID) || defined(__ANDROID__))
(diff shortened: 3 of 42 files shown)


Release/src/http/client/http_client_asio.cpp
```diff
@@ -325,32 +325,53 @@ class asio_connection_pool
     boost::asio::io_service& m_io_service;
     const int m_timeout_secs;
     const bool m_start_with_ssl;
-    const std::function<void(boost::asio::ssl::context&)>& m_ssl_context_callback;
+    std::function<void(boost::asio::ssl::context&)> m_ssl_context_callback;
```
ras0219-msft: Why is this change needed?

Release/include/pplx/threadpool.h
```diff
@@ -156,6 +173,9 @@ class threadpool
     std::vector<pthread_t> m_threads;
     boost::asio::io_service m_service;
     boost::asio::io_service::work m_work;
+
+    std::mutex m_connection_pool_map_mutex;
+    std::map<std::string, std::shared_ptr<web::http::client::details::asio_connection_pool>> m_connection_pool_map;
```
ras0219-msft: When does this map ever get cleared?

Release/src/http/client/http_client_asio.cpp
```diff
@@ -325,32 +325,53 @@ class asio_connection_pool
...
-
 class asio_client : public _http_client_communicator
 {
 public:
     asio_client(http::uri address, http_client_config client_config)
     : _http_client_communicator(std::move(address), std::move(client_config))
-    , m_pool(crossplat::threadpool::shared_instance().service(),
-             base_uri().scheme() == "https" && !_http_client_communicator::client_config().proxy().is_specified(),
-             std::chrono::seconds(30), // Unused sockets are kept in pool for 30 seconds.
-             this->client_config().get_ssl_context_callback())
```
ras0219-msft: It seems that if I make multiple clients with the same host but different SSL context callbacks, the second client's callback will get dropped.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -325,32 +325,53 @@ class asio_connection_pool
...
+            if (proxy.credentials().is_set())
+            {
+                host.append(proxy.credentials().username());
+            }
+        }
+
+        m_pool = crossplat::threadpool::shared_instance().obtain_connection_pool(host, [this]()
+        {
+            return std::make_shared<asio_connection_pool>(crossplat::threadpool::shared_instance().service(),
+                base_uri().scheme() == "https" && !_http_client_communicator::client_config().proxy().is_specified(),
```
ras0219-msft: How does this handle different proxy settings for different `http_client`s going to the same base uri?

Release/include/pplx/threadpool.h
hanzhumsft: The map is get cleared when the program exits.  Although the connection pool will remain until the process exists, all actual connections in it have a default 30 seconds timer.  The timer is started after the connection being released, and the underlying socket is closed once the timer expires.

Using `std::weak_ptr` here will make the pool cleared right after the last `http_client` using it. If another `http_client` which is going to use the same pool is created after that, all connections have to be re-started.

Release/src/http/client/http_client_asio.cpp
hanzhumsft: This is to prevent a dangling reference, because the lifetime of `asio_connection_pool` can be longer than the function object `m_ssl_context_callback` references. That function object comes from `http_client_config` when creating a `http_client` object.

Release/src/http/client/http_client_asio.cpp
hanzhumsft: Good catch!  I would make the pool not added to the map if ssl_context_callback is used.

Release/src/http/client/http_client_asio.cpp
hanzhumsft: I added both proxy and credentials information as part of the `host` variable here.

hanzhumsft:
Running tests on Linux:

**** outside_tests:outside_cnn_dot_com FAILED ****

**** outside_tests:outside_google_dot_com FAILED ****

**** outside_tests:multiple_https_requests FAILED ****

**** outside_tests:reading_google_stream FAILED ****

Finished running all 777 tests.

All 4 cases failed without my change.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -325,32 +325,64 @@ class asio_connection_pool
...
+                  host.append(proxy.credentials().username());
+                }
+            }
+
+            m_pool = crossplat::threadpool::shared_instance().obtain_connection_pool(host, [this]()
+            {
+                return std::make_shared<asio_connection_pool>(crossplat::threadpool::shared_instance().service(),
+                    base_uri().scheme() == "https" && !_http_client_communicator::client_config().proxy().is_specified(),
+                    std::chrono::seconds(30), // Unused sockets are kept in pool for 30 seconds.
+                    this->client_config().get_ssl_context_callback());
```
ras0219-msft: This branch of the `if` already has guaranteed there is no ssl_context_callback, so passing in `nullptr` here is probably more clear.

Release/include/pplx/threadpool.h
```diff
@@ -97,6 +101,19 @@ class threadpool
         return m_service;
     }
 
+    std::shared_ptr<web::http::client::details::asio_connection_pool> obtain_connection_pool(const std::string &base_uri, std::function<std::shared_ptr<web::http::client::details::asio_connection_pool>()> connection_pool_generator)
```
ras0219-msft: The implementation for this function needs to go into `Release\src\pplx\threadpool.cpp` file instead of the header.

Since we're only using this function with a single callback, it seems simpler to me if the single callback was inlined into this function. Instead of taking the base_uri, this function could just take the `http_client_config` by reference and determine whether to make a new pool or to reuse a cached pool based on the `base_uri()` and `ssl_context_callback`.

Release/include/pplx/threadpool.h
ras0219-msft: I'd like to see one more public function added which clears the entire cache. I'm concerned about applications that act more like scanners or crawlers; these applications would basically have a memory leak for every host they connect to.

While finer-grained control of the cache will eventually be needed, I want to ensure we have a workaround available until we can capture some solid cases where the optimal control design becomes clearer.

Release/include/cpprest/details/http_server_asio.h
```diff
@@ -87,7 +87,10 @@ class connection
         if (is_https)
         {
             m_ssl_context = utility::details::make_unique<boost::asio::ssl::context>(boost::asio::ssl::context::sslv23);
-            ssl_context_callback(*m_ssl_context);
+            if (ssl_context_callback)
```
hanzhumsft: It is possible that the `ssl_context_callback` does not have a valid target.

Release/include/cpprest/http_client.h
```diff
@@ -100,7 +100,6 @@ class http_client_config
 #endif
         , m_set_user_nativehandle_options([](native_handle)->void{})
 #if !defined(_WIN32) && !defined(__cplusplus_winrt)
-        , m_ssl_context_callback([](boost::asio::ssl::context&)->void{})
```
hanzhumsft: If we initialize `m_ssl_context_callback` with an do-nothing lambda, then `http_client_config::get_ssl_context_callback()` always returns a `std::function` that has a valid target.  As the target of `std::function` is hard to compare, we are unable to reuse the connection pool.

Release/include/pplx/threadpool.h
hanzhumsft: A pool timer is added to the `asio_connection_pool`.  The pool is going to be cleared 60 seconds after the last `http_client` using it releases it.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -80,7 +80,7 @@ class asio_connection
     asio_connection(boost::asio::io_service& io_service, bool start_with_ssl, const std::function<void(boost::asio::ssl::context&)>& ssl_context_callback) :
     m_socket(io_service),
     m_ssl_context_callback(ssl_context_callback),
-    m_pool_timer(io_service),
+    m_connection_timer(io_service),
```
hanzhumsft: Rename it to `m_connection_timer` because the timer is for `asio_connection`.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -252,7 +255,10 @@ class asio_connection_pool
     m_io_service(io_service),
     m_timeout_secs(static_cast<int>(idle_timeout.count())),
     m_start_with_ssl(start_with_ssl),
-    m_ssl_context_callback(ssl_context_callback)
+    m_ssl_context_callback(ssl_context_callback),
+    m_pool_timeout_secs(60), // Clean this connection pool 60 secs after the last asio_client release it.
+    m_pool_timer(io_service),
+    m_use_count(0)
```
ras0219-msft: Manual reference counting is extremely error prone; I understand that to accomplish the exact cleanup design you've specified (60 seconds after last asio_client) it seems needed, but this alone is a reason to change the cleanup design.

Instead of having a separate timer, what about having the pool automatically cleaned up when the last connection times out? That should be as simple as "whenever removing a connection from a connection_pool, check if the set of connections for this endpoint is now empty. If so, remove the entry for this endpoint".

hanzhumsft: But it's possible that at that point some http_client objects still hold the pool, although all individual connections have closed.  In this case we cannot clean up the entry.  I agree that manual reference counting is error-prone, so in the new change I removed the manual reference counting, and rely on shared_ptr's use_count.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -354,9 +370,49 @@ class asio_connection_pool
...
+        pool->start_pool_timer(boost::bind(&asio_connection_pool_map::free_connection_pool, this, boost::asio::placeholders::error, pool_key));
+    }
+}
+
+void asio_connection_pool_map::free_connection_pool(const boost::system::error_code &ec, const std::string &pool_key)
+{
+    if (!ec)
+    {
+        std::lock_guard<std::mutex> lg(m_connection_pool_map_mutex);
+        auto &pool = m_connection_pool_map[pool_key];
```
ras0219-msft: This will allocate a new pool `shared_ptr` if there is not a map entry. It would probably be better to use `.find(pool_key)` to avoid allocating a new map entry just to erase it.

hanzhumsft: Good point.

hanzhumsft:
Instead of having a "pool of pools" and a timer for each `asio_connection_pool`, I made the existing `asio_connection_pool` have dual functionalities.  With a `std::multimap` it can serve as a global pool shared by all `asio_client`.  With a `std::set`, it also can be a private pool for just one `asio_client` if the client has `ssl_context_callback` defined.

PR #172, opened 2016-06-04
Making oauth2_config::_request_token protected

deeringc:
To allow library users to subclass oauth2_config to implement custom oauth extensions

### DIFF
--- a/Release/include/cpprest/oauth2.h
+++ b/Release/include/cpprest/oauth2.h
@@ -466,6 +466,9 @@ public:
 		m_proxy = proxy;
 	}
 
+protected:
+	_ASYNCRTIMP pplx::task<void> _request_token(uri_builder& request_body);
+
 private:
     friend class web::http::client::http_client_config;
     friend class web::http::oauth2::details::oauth2_handler;
@@ -475,8 +478,6 @@ private:
         m_bearer_auth(true),
         m_http_basic_auth(true)
     {}
-
-    _ASYNCRTIMP pplx::task<void> _request_token(uri_builder& request_body);
 
     oauth2_token _parse_token_from_json(const json::value& token_json);
 


ras0219-msft:
Can you go into a bit more detail about what custom oauth extensions you're hoping to add with this change?

Inheritance is a very tightly-coupling extension mechanism, so I'd only like to use this approach if it cannot be accomplished in another way.

deeringc:
We have a proprietary grant type that operates very similarly to oauth2_config::token_from_code but with a different grant_type and different query arguments.

I would happily contribute this back as a concrete implementation but it's not of any use to anyone else. I guess others will encounter a need to extend the standard oauth flows so an extension point is worthwhile in my opinion.

I guess we could model this as a new API that takes a first argument of grant_type and a second argument of vector of pairs of string_t, then builds up a uri with those values and then calls _request_token.

Something along the lines of:

``
	pplx::task<void> custom_request(utility::string_t custom_grant_type, const std::vector<std::pair<utility::string_t>> custom_query_args )
    {
        uri_builder ub;
        ub.append_query(details::oauth2_strings::grant_type, custom_grant_type, false);
		
		for( const auto& arg : custom_query_args)
		{
			ub.append_query(arg.first, arg.scond);
		}
	  
        return _request_token(ub);
    }
``

ras0219-msft:
Would making `_request_token` into a public API also work for you?

deeringc:
Yes that works.

ras0219-msft:
As part of our 3.0 release, I would like to improve the OAuth2 APIs so we can move them out of `experimental`. I added support for Extension Grants as part of the changes, which I think will work for your case. In addition, I would appreciate any other feedback you have on the API changes.

The Extension Grant API is at
https://github.com/Microsoft/cpprestsdk/blob/192541718bdae3298cfcc06bd17d11d22a8cbbb1/Release/include/cpprest/oauth2.h#L453

I've also updated the OAuth2Client sample to use the new APIs
https://github.com/Microsoft/cpprestsdk/blob/192541718bdae3298cfcc06bd17d11d22a8cbbb1/Release/samples/Oauth2Client/Oauth2Client.cpp

PR #200, opened 2016-07-13
Fixed compilation errors on the latest OS X(El Capitan)

ixSci:
Clang generates errors on warnings complaining about redundant use of std::move. So
the redundant calls were removed.


ras0219-msft:
Thank you for the PR. Please reopen it against the development branch.

PR #214, opened 2016-07-30
Adding PARENT_SCOPE

KrishnaPG:
To make it possible to include as sub-project from external projects. This lets the variable `Casablanca_LIBRARIES` be accessible from the parent project


KrishnaPG:
Currently it is not possible to access the Casablanca output library name from a parent project. With this change, parent projects can access the output library name.

ras0219-msft:
This looks great. Thank you for the contribution!

PR #216, opened 2016-08-02
Allow const access operator for json objects and arrays

pdgendt:
Const access operators would shorten some conditional checks.

For example:

```C++
void some_function(const web::json::value& obj)
{
    if (obj.has_field("someField") && obj.at("someField").is_string()) {
        // do something
    }
}
```
Would become:
```C++
void some_function(const web::json::value& obj)
{
    if (obj["someField"].is_string()) {
        // do something
    }
}
```

Also readable chaining:
```C++
void some_other_function(const web::json::value& obj)
{
    if (obj["someObject"]["someNestedObject"]["someField"].is_string()) {
        // do something
    }
}
```

### DIFF
--- a/Release/include/cpprest/json.h
+++ b/Release/include/cpprest/json.h
@@ -220,6 +220,12 @@ namespace json
         /// <returns>The JSON value object that contains the result of the assignment.</returns>
         _ASYNCRTIMP value &operator=(value &&) CPPREST_NOEXCEPT ;
 
+        /// <summary>
+        /// Returns the static immutable null reference
+        /// </summary>
+        /// <returns>A static null reference</returns>
+        static _ASYNCRTIMP const value& __cdecl null_ref();
+
         // Static factories
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ras0219-msft:
I really like how literate the operations become, however the differing behavior from the non-const overloads (non-const overloads currently will transform null values into the "correct" kind) prevents me from accepting this PR. Unfortunately, the behavior that the non-const `operator[]` overloads is impossible to implement in const overloads.

I do think there could be value in some new accessor function (at cost of being slightly more verbose) that does have the behavior you're proposing here. For example, perhaps `at_opt()`:
```
auto&& v = obj.at_opt("someObject").at_opt("somePlace").at_opt(5);
if (v.is_number()) { ..... }
```
I agree that your proposed APIs are more fluent, but its behavior conflicts with the other existing APIs.

I'll close this PR for now, but I'd be happy to continue discussing this here or in a new PR.

PR #217, opened 2016-08-02
Fixed race condition in Casablanca ws_client_winrt.cpp

daniuk-microsoft:
Fixed race condition in Casablanca ws_client_winrt.cpp
- Simplified message queuing logic to always use the outgoing queue
- Removed m_num_sends member variable

This addresses a race condition between m_num_send and m_outgoing_msg_queue. m_num_send can be incremented twice, but if the 2nd caller is preempted before pushing msg into m_outgoing_msg_queue, then a concurrent execution of send_msg() can call m_outgoing_msg_queue.front(). This results in an access violation.  This modification removes m_num_send entirely and only relies on m_outgoing_msg_queue to keep track of pending messages (the front of the queue is the message in-flight)


kavyako:
This change will be a part of the 2.9 release. Thank you for contributing.

PR #218, opened 2016-08-03
Fixed mem leak in HttpRequestCallback::OnHeadersAvailable()

daniuk-microsoft:
IXMLHTTPRequest2::GetAllResponseHeaders() allocates memory that is
returned via hdrStr. Change to make sure this gets deallocated.


daniuk-microsoft:
Hello, could I get this change reviewed for merge? This would be an important fix to get into the release build.

ras0219-msft:
Sorry about the delay, unfortunately we were already in the process of releasing and were unable to accept this change in time for 2.9.

On the other hand, this is a really solid fix! Thank you for contributing it and it will certainly be in the next release.

PR #219, opened 2016-08-04
Fix namespace issue with proxy_tests

emonty:
credentials with no namespace is ambiguous when built on Fedora with
krb5 things installed. Being explicit that we want web::credentials here
and not the credentials struct from krb5.h solves the issue.


ras0219-msft:
Thank you for the PR; Could you make it out against the development branch? I believe this issue was already solved there.

PR #221, opened 2016-08-06
wish Rest SDK C++ is going to support Http/2

abcdef123ghi:
I wish 3.0.0 is going to support HTTP/2


ras0219-msft:
Hi, this looks like it should be an issue instead of a Pull Request since there is no code to be contributed. Please feel free to reopen this as an issue.

PR #233, opened 2016-09-07
Http compression

deeringc:
Implementing HTTP compression for Windows only initially. There are a few things left for Windows support but I wanted to get your feedback before I go any further.


deeringc:
For now, I'm using Nuget for the zlib dependency on Windows. What would you like me to use when I implement the same for OSX?

Release/src/build/vs14.wod/casablanca140.wod.vcxproj
```diff
@@ -21,14 +21,17 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.root))\Build\Release.Product.Settings" />
   <ImportGroup Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+    <Import Project="..\..\..\..\packages\zlib.v140.windesktop.msvcstl.static.rt-dyn.1.2.8.8\build\native\zlib.v140.windesktop.msvcstl.static.rt-dyn.targets" Condition="Exists('..\..\..\..\packages\zlib.v140.windesktop.msvcstl.static.rt-dyn.1.2.8.8\build\native\zlib.v140.windesktop.msvcstl.static.rt-dyn.targets')" />
```
deeringc: I'm not entirely sure what this "wod.vcxproj" project is, but it seemed like I needed to add a nuget reference here to get it to build.

Release/src/http/client/http_client_winhttp.cpp
```diff
@@ -1159,12 +1184,21 @@ class winhttp_client : public _http_client_communicator
...
 
                     if(num_bytes > 0)
                     {
-                        auto writebuf = p_request_context->_get_writebuffer();
-                        p_request_context->allocate_reply_space(writebuf.alloc(num_bytes), num_bytes);
+                        if (p_request_context->decompressor)
+                        {
+                            // Decompression is too slow to reliably do on this callback. Therefore we need to store it now in order to decompress it at a later stage in the flow.
+                            // However, we want to eventually use the writebuf to store the decompressed body. Therefore we'll store the compressed body as an internal allocation in the request_context
+                            p_request_context->allocate_reply_space(nullptr, num_bytes);
```
deeringc: When I tried to do the decompression here, I ran into an issue where a another callback to WINHTTP_CALLBACK_STATUS_READ_COMPLETE seemed to come in before the decompression had completed, and before that decompressed data was available. 

I'm now doing the decompression in the WINHTTP_CALLBACK_STATUS_READ_COMPLETE callback instead.

deeringc: Here we're storing the compressed body in the internal allocation of the p_request_context. 

Once we decompress later, the decompressed data will be stored in the writebuf, and will be added as the external allocation in the p_request_context.

Release/src/http/client/http_client_winhttp.cpp
```diff
@@ -1220,9 +1254,39 @@ class winhttp_client : public _http_client_communicator
...
+                        }
+
+                        // We've decompressed this chunk of the body, need to now store it in the writebuffer.
+                        auto decompressed_size = decompressed.size();
+
+                        if (decompressed_size > 0)
+                        {
+                            auto p = writebuf.alloc(decompressed_size);
+                            p_request_context->allocate_reply_space(p, decompressed_size);
+                            std::memcpy(p_request_context->m_body_data.get(), &decompressed[0], decompressed_size);
```
deeringc: So, I'm not happy with this copy, but the difficulty is that we cannot know the size of the decompressed data without actually decompressing the data, and to allocate in the writebuf we need to know the size. 

Any ideas?

Release/src/http/common/http_helpers.cpp
```diff
@@ -427,5 +431,310 @@ bool validate_method(const utility::string_t& method)
...
+                }
+
+                input_offset = read_output(input_offset, available_input, total_out_before, temp_buffer, output);
+            }
+
+            return output;
+        }
+    };
+
+    class stream_compressor::stream_compressor_impl : public compression_base_impl
```
deeringc: This isnt actually used in production code (yet). Sending compressed content in a request isn't strictly standardised. There was a request in the github issue to support this however. It is also needed for one of the tests below.

Release/src/http/common/http_helpers.cpp
```diff
@@ -427,5 +431,310 @@ bool validate_method(const utility::string_t& method)
...
+            return output;
+        }
+
+        ~stream_compressor_impl()
+        {
+            deflateEnd(&stream());
+        }
+    };
+#else // Stub impl for when compression is not supported
+
```
deeringc: In order to avoid #ifdefs all over the code, I implemented a stub in the cases where http compression isnt supported.

Release/src/build/vs14.wod/casablanca140.wod.vcxproj
ras0219-msft: We should delete the wod project. It was for the Intel Galileo, however IoT core pretty much supersedes it in every way and is covered by UWP. I believe we've already removed it from the "dirs.proj" build tree.

Release/src/http/client/http_client_winhttp.cpp
ras0219-msft: My best idea would be to take a guess and then fixup if we're wrong. I think this is fine for a first implementation though.

Release/src/http/common/http_helpers.cpp
ras0219-msft: This should probably throw errors instead of "doing nothing".

ras0219-msft:
We rely on homebrew for our dependencies on OSX, but you can just use CMake's "find_package(ZLIB)" directive to get zlib on non-Windows.

Overall, this looks great. I'll be happy to merge it once you're done 👍 .

Release/src/http/client/http_client_asio.cpp
```diff
@@ -1316,25 +1346,78 @@ class asio_context : public request_context, public std::enable_shared_from_this
         {
             // more data need to be read
             const auto this_request = shared_from_this();
-            writeBuffer.putn_nocopy(boost::asio::buffer_cast<const uint8_t *>(m_body_buf.data()),
-                             static_cast<size_t>(std::min(static_cast<uint64_t>(m_body_buf.size()), m_content_length - m_downloaded)))
```
deeringc: One think I'm not totally clear about is the difference between the result of this std::min evaluation and the subsequent:

`writtenSize = op.get();`

size. Could you elaborate a little on what's happening?

Release/src/http/common/http_helpers.cpp
deeringc: Yeah, sounds like a better idea!

Release/tests/functional/http/client/outside_tests.cpp
```diff
@@ -63,6 +63,27 @@ TEST_FIXTURE(uri_address, outside_cnn_dot_com)
     });
 }
 
+TEST_FIXTURE(uri_address, outside_wikipedia_compressed_http_response)
```
deeringc: Changed this to wiki instead of google. Reason being that google is redirecting (maybe because I'm writing this from Europe with a different TLD) with a 302. That is handled fine on Windows (inside WinHTTP) but not on OSX (requires application level code to handle the redirect). Wiki isn't redirecting so it's a better test url. 

I will add a separate PR to fix up other tests that have similar issues.

Release/src/http/common/http_helpers.cpp
```diff
@@ -427,5 +433,294 @@ bool validate_method(const utility::string_t& method)
...
+        }
+    };
+
+    class stream_compressor::stream_compressor_impl : public compression_base_impl
+    {
+    public:
+        stream_compressor_impl(compression_algorithm) {}
+        compression::data_buffer compress(const uint8_t*, size_t, bool)
+        {
+            assert(false); //Compression is not implemented or configured on this platform
```
deeringc: So, I had this throwing an exception instead but LTCG detected the exception and that code that came after it was unreachable. This caused a warning which was treated as an error.

Release/tests/functional/http/client/outside_tests.cpp
ras0219-msft: Thank you, this is a much needed change. We've had a longstanding issue with this test failing outside the US.

Release/src/http/common/http_helpers.cpp
ras0219-msft: I think the right thing to do is to fix the caller appropriately, even if that means `ifdef`s.

If we just `return {}`, this code will certainly result in bug reports about "When I use compression on platform X all my data disappears!" or worse. Additionally, I think `assert(false)` _may_ be a bit strong here since I assume this datapath can be triggered by external input; e.g. if the website i'm pulling from suddenly decides to enable compression, my application hard crashes.

Release/src/http/client/http_client_asio.cpp
ras0219-msft: I believe the interface for `writeBuffer.putn_nocopy()` has provisions for potentially writing less data than requested. For example, if the underlying stream is a producer/consumer and has a smaller internal buffer than the provided output size. It may copy its entire buffer, then immediately complete the output task to continue working while it buffers more.

Release/src/http/common/http_helpers.cpp
deeringc: So all uses of this code are guarded by is_supported() so this should never actually be run, even if the server decides to send a compressed response unsolicited. In my mind if it does, it's a programmer error and we need to know about it in debug builds.

How about just returning the input rather than empty to address your data loss concerns?

Release/src/http/client/http_client_asio.cpp
deeringc: Ok, thanks. Is it safe for me to use the result of std::min to determine the size of the input buffer for decompression?

Release/src/http/common/http_helpers.cpp
```diff
@@ -427,5 +433,308 @@ bool validate_method(const utility::string_t& method)
...
+    };
+
+    class stream_decompressor::stream_decompressor_impl : public compression_base_impl
+    {
+    public:
+        stream_decompressor_impl(compression_algorithm) {}
+        compression::data_buffer decompress(const uint8_t* data, size_t size) 
+        {
+            //Decompression is not implemented or configured on this platform
+            if (!stream_decompressor::is_supported())
```
deeringc: Adding this conditional here seems to trick the LTCG into not seeing code that gets linked in after it as unreachable.

deeringc:
Hi @ras0219-msft. Thanks for the feedback so far! Please let me know what I should do so we can close out this PR

ras0219-msft:
First, sorry for the extreme delay. We put this on the back burner while making the 2.9 release, which took far longer than expected.

I've merged this in locally and I'm running our test suite, if everything looks good I'll push this to master soon. I made a few changes locally that will also be merged:
- The `outside_wikipedia_compressed_http_response` test needs to be skipped on platforms that don't have compression support compiled in.
- After some internal discussion, your intial implementation was a better choice: we should simply silently do nothing if compression is not available on that platform. This has the same effect as the server choosing to not opt-in, so from the client's perspective they can't see the difference. I reverted the `throw` conditionals, which also removes the annoying LTCG warnings.

deeringc:
No problem at all. Deadlines come first! Great news that it's being merged into mainline. 

Just a note, the only reason that I didn't enable compression support for other platforms is that Mac and Win are my current focus, and to a lesser degree I don't have access to the right build environments (eg. Android). For anyone looking to add support to Linux, Android etc.. the actual code that I added should be totally cross platform (based on the asio http client). All you'll need to do is modify the cmake script to add the zlib dependency and to add a few #ifdefs to enable the code for the given platforms.

PR #246, opened 2016-09-20
Enable TLS 1.1 and 1.2

nilfrisk:
Enable TLS 1.1 and 1.2 … 

By default only SSL3 and TLS1 are enabled in Windows 7 and Windows 8.
By default only SSL3, TLS1.0, TLS1.1, and TLS1.2 are enabled in Windows 8.1 and Windows 10.

https://msdn.microsoft.com/en-us/library/windows/desktop/aa384066


ras0219-msft:
Thanks for the PR!

Issue #249, opened 2016-09-26
Converting utf16 to utf8 costs a lot time


JasonYang-MSFT:
As mentioned in the title, utf16_to_utf8 costs a lot time. For 1 MB utf16 string, it would costs 5 seconds on Win10 with E5-1620 machine. And the costs increased by linear, 11 seconds for 2 MB, 22 seconds for 4 MB e.g.

Here is the sample code:

```
#include "stdafx.h"
#include "cpprest/asyncrt_utils.h"

int main()
{
	int length = 1 * 1024 * 1024;
	utility::datetime start = utility::datetime::utc_now();
	utility::string_t text(length, _XPLATSTR('C'));
	auto content = utility::conversions::utf16_to_utf8(text);
	utility::datetime end = utility::datetime::utc_now();
	ucout << "String Length: " << length << std::endl;
	ucout << "Time Cost: " << end - start << " second(s)" << std::endl;
    return 0;
}
```

grahamreeds:
This is minimally related to #72.

JasonYang-MSFT:
#72 was discussing about to avoid such kind of conversion. But I am more concerned about the conversion performance here.

grahamreeds:
The fastest code is code that isn't ran, and I did say it was kind of
related. Fixes in this area would help both.

PR #251, opened 2016-09-30
Improve the performance of conversion between UTF16 and UTF8 strings.

JasonYang-MSFT:
Related to #249 
Original costs for 8MB's conversion from utf16 to utf8 is 49s.
And after improvements for 1GB conversion only needs 1~2 seconds for ASCII data, 2~3 seconds for non-ASCII data.

Mainly improved by:
1. Count the length of the resulting sequence to avoid multiple memory allocation through push_back.
2. Use pointers instead of operator[]
3. Use index instead of iterator.

### DIFF
--- a/Release/src/utilities/asyncrt_utils.cpp
+++ b/Release/src/utilities/asyncrt_utils.cpp
@@ -377,12 +377,13 @@ std::string __cdecl conversions::utf16_to_utf8(const utf16string &w)
         // Check for high surrogate.
         if (src >= H_SURROGATE_START && src <= H_SURROGATE_END)
         {
-            const auto highSurrogate = w[++index];
+            const auto highSurrogate = src;
+			++index;
             if (index == w.size())
             {
                 throw std::range_error("UTF-16 string is missing low surrogate");
             }
-            const auto lowSurrogate = src;
+            const auto lowSurrogate = w[index];
             if (lowSurrogate < L_SURROGATE_START || lowSurrogate > L_SURROGATE_END)
             {
                 throw std::range_error("UTF-16 string has invalid low surrogate");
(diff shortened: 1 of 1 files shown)


grahamreeds:
How much extra performance is gained through using 1, 2, and 3?

I think 1 will give most benefit but I can't imagine 2 & 3 giving much.

Also pity the diff algorithm mangles the output.

JasonYang-MSFT:
The performance improved from (8MB, 49 seconds) to (1GB, 1~3 seconds).

2&3 are also import parts. With only 3, the performance improved from (8MB, 49 seconds) to (*MB, 3~4 seconds).
Also you can think about this: for 1GB data, the "for" statement will be running for more than one billion times. Save any instructions inside it would save billions of instructions.

Besides, Use pointers instead of operator[] would reduce the extra checking inside operator[].

grahamreeds:
I guess it makes sense.

Did you look at using `std::codecvt_utf8_utf16`?

I was thinking something along the lines of

```
std::u16string source; 
std::wstring_convert<std::codecvt_utf8_utf16<char16_t>,char16_t> convert; 
std::string dest = convert.to_bytes(source);   
```  

but I am not working with cpprestsdk any more (different firm) so I don't have much time to devote to issues (though still track it).

thedrow:
If you are trying to improve conversion performance than SIMD is a candidate for a huge boost in performance, at least when converting UTF-8 to UTF-16.
See https://woboq.com/blog/utf-8-processing-using-simd.html and https://github.com/woboq/utf8sse4/
Also http://u8u16.costar.sfu.ca/export/46/trunk/src/libu8u16.pdf is a useful paper on the subject.

As for counting the length, take a look at http://www.daemonology.net/blog/2008-06-05-faster-utf8-strlen.html
That article may give us a hint on how to devise an algorithm for counting the intended UTF-16 string length.

P.S.
I haven't really dived into the code. I'm just dropping my 2 cents here.

JasonYang-MSFT:
@thedrow 
Thanks for the reference.

ras0219-msft:
Thanks for the PR!

PR #257, opened 2016-10-06
bug fix : m_close initialization in wrong place

johnpaultaken:
m_close getting initialised in handle_headers wipes out the value assigned to it in handle_http_line.
This could cause some connection objects to not get released, for example if the request is HTTP1.0.

@ras0219-msft


johnpaultaken:
Please ignore the second commit. It is submitted as another patch (the next one).

ras0219-msft:
Thanks for this! Merged in 8518210 on master.

PR #259, opened 2016-10-07
fix for bug: connection objects are not getting released for some cases

johnpaultaken:
**fix for bug**: connection objects and their associated buffers don't get released in some cases.
In several places when errors happen during the request sequence the reference on the connection object is not getting decremented.
This causes memory leaking of connection objects and their associated boost asio buffers.
This is most serious at location in handle_http_line() where while waiting for the next HTTP request from peer, the server can get a connection reset by peer quite frequently resulting in significant memory bloating.
**Following elaborates the leaking sequence**:
On certain request error conditions, the current code is doing m_close=true followed by a do_response() and returns from the request sequence.
But before ending the request sequence the code must decrement the reference count on connection object initialized to 1 at construction.
do_response() begins a response sequence after incrementing the reference count which is decremented in finish_request_response.
But the reference count from construction remains and the connection object never gets released. 
So the fix is to decrement the reference count before returning from the request sequence.
@ras0219-msft

### DIFF
--- a/Release/src/http/listener/http_server_asio.cpp
+++ b/Release/src/http/listener/http_server_asio.cpp
@@ -258,6 +258,11 @@ void connection::handle_http_line(const boost::system::error_code& ec)
             m_request._reply_if_not_already(status_codes::BadRequest);
             m_close = true;
             do_response(true);
+            if (--m_refs == 0)
+            {
+                delete this;
+            }
+            return;
         }
     }
     else
@@ -285,6 +290,10 @@ void connection::handle_http_line(const boost::system::error_code& ec)
             m_request.reply(status_codes::BadRequest);
             m_close = true;
             do_response(true);
+            if (--m_refs == 0)
+            {
+                delete this;
+            }
             return;
         }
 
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thanks for the PR!

In order to feel more certain that we've caught _all_ of the cases, I added dummy types to document the actual "contracts" of the async functions in 38d916a638. For example:

```c++
will_erase_from_parent_t do_bad_response()
{
    ++m_refs;
    m_request.get_response().then([=](pplx::task<http::http_response> r_task)
    {
        http::http_response response;
        try
        {
            response = r_task.get();
        }
        catch (...)
        {
            response = http::http_response(status_codes::InternalError);
        }
        // before sending response, the full incoming message need to be processed.
        serialize_headers(response);
        (will_deref_and_erase_t)async_write(&asio_server_connection::handle_headers_written, response);
    });
    return will_erase_from_parent_t{};
}
```

This hopefully makes it simply and statically documented that
1. `do_response()` is responsible for eventually calling erase.
2. `do_response()` is _not_ responsible for consuming a refcount.
3. Internally, `do_response()` increments the refcount and asynchronously forwards to a function responsible for consuming a refcount _and_ erasing.

PR #260, opened 2016-10-07
ECONNRESET must be used instead of ECONNABORTED.

johnpaultaken:
fix for bug: when client does not perform an orderly shutdown of the connection during recv() or read() the error code returned is 'connection reset by peer' or 'connection timed out' according to the open group specification (also linux specification). It is not 'operation aborted' which the current code assumes. 'operation aborted' happens during accept().
This fix was needed on RHEL 7 server.
References:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/recv.html
https://linux.die.net/man/3/recv

### DIFF
--- a/Release/src/http/listener/http_server_asio.cpp
+++ b/Release/src/http/listener/http_server_asio.cpp
@@ -249,7 +249,12 @@ void connection::handle_http_line(const boost::system::error_code& ec)
     if (ec)
     {
         // client closed connection
-        if (ec == boost::asio::error::eof || ec == boost::asio::error::operation_aborted)
+        if (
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thanks for the fix and references. Are any of the other async callbacks affected by this issue?

johnpaultaken:
@ras0219-msft   Thank you for your review. I was investigating an issue where our server was growing memory usage gradually such that it had to be restarted every week. I found three issues while investigating that issue which is submitted as patches #257, #259 and this one. Since these fixed our memory growth problem, I didn't look at other parts of the code to see any other parts may have the same issue. I can do that some time later when I have free time.

PR #305, opened 2016-12-05
Enable IPv6 addresses to be resolved and specified as valid URI

gwishnie:
An IPv6 address must be enclosed in [] when in a URI (i.e.
http://[:::::::]:80).  This change allows the '[' and ']' as characters
allowed as part of the authority so the underlying HTTP library can
receive a resolved IPv6 address and utilize.  The subsystem itself must
support utilizing, this change only allows it to be treated as valid.


ras0219-msft:
This looks pretty good, thanks for also adding tests 👍 .

If you've signed the CLA but the bot isn't updating, you may need to close this PR and open a new one.

gwishnie:
Thanks, I did sign.  Will close/resubmit.

PR #307, opened 2016-12-06
Suppress SSL leak functions that aren't present in LibreSSL

peitschie:


ras0219-msft:
Thanks for the resubmission.

I renamed the macro to avoid using _CAPITAL, since those macro names are reserved. We're pretty bad about using them in a lot of places currently, but that's not a reason to not improve :).

PR #309, opened 2016-12-07
Avoid potential buffer overrun

cygnus9:
The compiler is allowed to read three bytes regardless of the fields that are accessed. So if you're last valid byte is the last byte of a memory page this may trigger a segfault.


ras0219-msft:
Thanks, this is a really nasty bug!

PR #311, opened 2016-12-08
Include ZLIB dirs when websockets are active

peitschie:
Tiny little fixup required on Windows when using an openssl library that doesn't have zlib in the folder.


ras0219-msft:
Thanks!

PR #312, opened 2016-12-14
add *.hpp and *.dat HEADERS_DETAILS to CMake file globbing

hrantzsch:
make install fails to install include/cpprest/details/{http_constants.dat,SafeInt3.hpp}, as *.hpp and *.dat are not globbed from the respective folder. As a result, the installed library cannot be used.


ras0219-msft:
Thanks! I must have regressed this during the last CMake refactoring.

PR #319, opened 2016-12-27
Solution does not compile in folders with spaces in the path

HolgerStrauss:
For the cpprestsdk140.uwp project, the paths must be quoted in the Post-Build Event.


ras0219-msft:
Thanks!

PR #336, opened 2017-01-19
Don't add default port to 'Host' header in ASIO

reneme:
If requesting data through a pre-signed S3 URL [1], the signature calculation includes the hostname provided in the 'Host' header. Appending the default port number - as done in `asio_context::start_request()` - invalidates this signature and prevents us from down-/uploading data through such URLs.

Since appending the port number to the 'Host' header is optional [2], we can omit it in the default case. This fixes the usage of pre-signed S3 URLs with C++ REST SDK.

A work-around for this fix is setting the 'Host' port appropriately in the `http_request` object when making a request through `http_client`.

Note that this works perfectly fine on Windows when using the *WinHTTP* backend.

[1] http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html
[2] https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

### DIFF
--- a/Release/src/http/client/http_client_asio.cpp
+++ b/Release/src/http/client/http_client_asio.cpp
@@ -659,7 +659,11 @@ public:
             // Add the Host header if user has not specified it explicitly
             if (!ctx->m_request.headers().has(header_names::host))
             {
-                request_stream << "Host: " << host << ":" << port << CRLF;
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thanks!

PR #365, opened 2017-02-22
Use 1.0.2k openssl library - Play store rejects 1.0.2d

pnemere:
On Android we need a newer openssl library otherwise this nuget package is pretty much useless. Play store scans and somehow finds compiled 1.0.2d in binary and rejects the upload.


ras0219-msft:
Thanks!

Issue #171, opened 2016-06-03
[question] howto disable redirect?


petke:
Hi

I GET a url that redirects (302) to a new url. cpprestsdk automatically redirects to the new url. In this case that's not what I want. I want to disable redirection, so that I can get at the http response header of the original url. (This header has important Set-Cookie information. I need to store this cookie in my application, so that I can use this cookie later for other GET calls)

How do I do this?

petke:
[bump]

petke:
[bump]

ras0219-msft:
You will need to use `http_client_config::set_nativehandle_options()` to set the WINHTTP_OPTION_REDIRECT_POLICY:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa384066(v=vs.85).aspx

zwaheed-rms:
I was able to disable redirects on windows by using the code shown below, does anyone know how to achieve the same on Linux?

```
void CallbackFunc(native_handle handle) {
  // disable auto redirects
  DWORD dwOptionValue = WINHTTP_DISABLE_REDIRECTS;
  BOOL sts = WinHttpSetOption(handle, WINHTTP_OPTION_DISABLE_FEATURE, &dwOptionValue, sizeof(dwOptionValue));
  if (!sts) {
    std::cerr << "WINHTTP_DISABLE_REDIRECTS failed.\n";
  }
}

// later create http_client like shown below 
http_client_config config;
config.set_nativehandle_options(CallbackFunc);
http_client client(U("http://some_url/"), config);
```

deeringc:
@zwaheed-rms, I don't believe that redirects are implemented in the asio-based http client which is used on Linux.

drearyworlds:
Necro-posting just to show a more concise way to do it that worked for me:
```
        http_client_config clientConfig;
        clientConfig.set_nativehandle_options([&](web::http::client::native_handle nativeHandle) {
            // Disable auto redirects
            DWORD dwOptionValue = WINHTTP_DISABLE_REDIRECTS;
            BOOL sts = WinHttpSetOption(nativeHandle, WINHTTP_OPTION_DISABLE_FEATURE, &dwOptionValue, sizeof(dwOptionValue));
            if (!sts) {
                std::cerr << "WINHTTP_DISABLE_REDIRECTS failed.\n";
            }
        });
```

PR #373, opened 2017-03-02
Add redirect support for ASIO client

8W9aG:
* Add a checking task that determines whether to
launch a new request with the redirection
* Adheres to RFC-2068 10.3 by not allowing more than
5 redirects
* Does not allow redirect loops

Let me know if this approach is sound, if so then I will make tests.


8W9aG:
@stgates would it be possible to give this a little looksee?

mobileben:
Just curious on this. What happens for any exceptions generated?  I've found that when using the ```.then```, exceptions can't be caught. I haven't overly looked at the code for pplx. What I've had to do in my stuff is wrap the call in ```std::async``` in order to prevent a ```SIGTRAP```. For example, a simple network error, if you don't do this, will crash the client due to an uncaught exception.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -1659,7 +1663,66 @@ pplx::task<http_response> asio_client::propagate(http_request request)
...
-    return result_task;
+    return result_task.then(std::bind(&asio_client::maybe_redirect,
+                                      this,
+                                      std::placeholders::_1,
+                                      request,
+                                      std::vector<uri>()));
+}
+
+pplx::task<http_response> asio_client::maybe_redirect(http_response response, http_request request, std::vector<uri> redirected_uris) {
+    static const status_code found_status_code = 302;
```
deeringc: This constant is already defined here: `web::http::status_codes::Found`

deeringc: Shouldnt we also react to:

- 301 Moved Permanently
- 307 Temporary Redirect
- 308 Permanent Redirect

garethsb-sony:
If this were to be added, it would be good to add to `http_client_config` a common setting that would enable/disable this behaviour in both ASIO and WinHttp implementations (for the latter, see https://github.com/microsoft/cpprestsdk/issues/171)

Release/src/http/client/http_client_asio.cpp
garethsb-sony: Yes. Also, the PR code comments refer to RFC 2068, which has been obsoleted twice-over, with [RFC 7231 section 6.4](https://tools.ietf.org/html/rfc7231#section-6.4) being the most recent spec I believe. This spec also removes the arbitrary maximum redirect count of 5 and explicitly recommends identifying cycles.

garethsb-sony:
Specifically, config support could be made common for:

* the maximum number of redirects permitted (see [WINHTTP_OPTION_MAX_HTTP_AUTOMATIC_REDIRECTS](https://docs.microsoft.com/en-gb/windows/win32/winhttp/option-flags?#WINHTTP_OPTION_MAX_HTTP_AUTOMATIC_REDIRECTS)) with zero implying no redirection, and a sensible default of 5 (as in now-obsoleted [RFC 2616 Section 10.3](https://tools.ietf.org/html/rfc2616#section-10.3) or 10 (as WinHttp)
* whether redirects from HTTPS to HTTP are permitted (see [WINHTTP_OPTION_REDIRECT_POLICY_DISALLOW_HTTPS_TO_HTTP](https://docs.microsoft.com/en-gb/windows/win32/winhttp/option-flags?#WINHTTP_OPTION_REDIRECT_POLICY))

PR #381, opened 2017-03-08
Do not link to boost includes publicly in static library

8W9aG:
* Static library builds do not need the have public
include directories (since they aren’t installed)
* This allows people building boost themselves from
source to include their own boost directories without
running into the error:
error INTERFACE_INCLUDE_DIRECTORIES property contains path
which is prefixed in the source directory


ras0219-msft:
Thanks for the PR!

I've overhauled our CMake buildsystem to better support installation.
Notably:
- Installation can be disabled with `-DCPPREST_INSTALL=OFF`
- We should no longer export absolute paths into our cmake config file, however if your cmake version is new enough to support imported targets for zlib/boost/openssl, we will correctly express dependencies upon them.

I believe this should address both of your concerns listed above -- please try out the install and let me know if I missed anything.

PR #422, opened 2017-04-21
Fixed non-conformant extra qualification on member function declarations

guaranteed-to-be-unique:


ras0219-msft:
Thanks for the PR!

I was unable to merge this directly due to missing the CLA, however because the fix is so small I've made the equivalent change in master. Thank you very much for the contribution!

PR #429, opened 2017-04-29
Change end-of-line encoding of two files to Unix (LF)

c72578:
The line encoding of the following two files is CR+LF. In order to
harmonize eol encoding with the other files, it is changed using:
dos2unix Release/src/http/oauth/oauth1.cpp
dos2unix Release/libs/websocketpp/websocketpp/sha1/sha1.hpp


ras0219-msft:
Thanks!

Issue #436, opened 2017-05-08
Building iOS doesn't seem to work


mobileben:
Okay this may be just plain old user error. I followed the steps, minus the building of boost and openssl, since I already have them pre-built (although boost is not a framework).

I am more or less doing these steps:

```
git clone https://github.com/cristeab/ios-cmake.git
pushd ios-cmake
git apply ../fix_ios_cmake_compiler.patch
popd
mkdir build.debug
pushd build.debug
cmake .. -DCMAKE_BUILD_TYPE=Debug
make
popd
```

Which was listed in the instructions. This generates errors.

Determining if the include file pthread.h exists failed with the following output:
Change Dir: /Users/me/Development/3rdParty/cpprestsdk/Build_iOS/build.debug/build.i386/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/xcodebuild" "-project" "CMAKE_TRY_COMPILE.xcodeproj" "build" "-target" "cmTC_22e4e" "-configuration" "Debug"
=== BUILD TARGET cmTC_22e4e OF PROJECT CMAKE_TRY_COMPILE WITH CONFIGURATION Debug ===

Check dependencies
target specifies product type 'com.apple.product-type.tool', but there's no such product type for the 'iphonesimulator' platform

** BUILD FAILED **

The following build commands failed:
	Check dependencies
(1 failure)

Both the ios-cmake and the contents of this Build_iOS directory seem a bit old. Is this even buildable for iOS anymore? While probably ill-advised, I was using ```http_client```, which is really what I want this lib for. Mainly because I have the server already using cpprestsdk and thought it would be more convenient if I could use the client too. Maybe though that is the wrong approach here.

mobileben:
So it looks like iOS.cmake needs another patch. It needs the following lines:

set(CMAKE_MACOSX_BUNDLE ON)
set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_REQUIRED OFF)

I got that from here.

https://gitlab.kitware.com/cmake/cmake/issues/16695

BTW, is there any reason the approach of cloning iOS.cmake is being done? It just seems to add steps. Since there is no guarantee that iOS.cmake will be updated or even maintained in a way where it is compatible, it would sort of seem better to make an attribution to the source but have the version in use in Build_iOS. Cloning and patching just seems a bit off as an approach.

I'll try and do a build a bit later. In the meantime, it did finally run through cmake after the changes. 

[...]

PR #439, opened 2017-05-17
Update cmake to build with Xcode8 as well as boost libraries instead …

mobileben:
…of framework (#436)


mobileben:
@ras0219-msft I'm in the process of updating to the latest of cpprestsdk. I wasn't aware of this conflict, but can fix it. I plan on closing this and opening a new pull request.  It will also contain some other changes that should make building for iOS more consistent (some recent issues flag not being able to build). 

Will changes like this get pulled in once I submit the pull request (obviously you have to review)? It would be good to have a more stable way for iOS users to onboard.

mobileben:
Replacing this PR with https://github.com/Microsoft/cpprestsdk/pull/776

Issue #428, opened 2017-04-27
http_listener default thread count - 40 threads


zerodefect:
After trying the cpprest library out (on Linux), specifically the http_listener, I have quickly discovered that the default number of threads that are created is 40 threads. Nonetheless, I've examined the  code, and I have been unable to ascertain why a value of 40 has been chosen.

A hard-coded thread count is likely to go against a popular school of thought that the number of threads should match the number of CPU cores (or at the most: thread_count <= (2 x cpu cores)).  For example, my development system contains 4 cores.

Having thought about this further, there should be two (2) potential changes - the first one is easier than the second:
1. The thread count should be configurable via a parameter during intialisation of the http_listener.
2. Applicable classes should enable developers to instantiate instances of an object specifying a custom thread pool.     

Environment:
Ubuntu Zesty 17.04
Boost 1.62
GCC 6.3
cpprest 2.9

zerodefect:
In _Release/src/pplx/threadpool.cpp_, I added the following code to detect/set the default thread count.

    namespace
    {
        
    inline size_t default_thread_count()
    {
        // Hardware_concurrency may return 0. Nonetheless, need at least two threads.
        return std::max<size_t>(size_t{2},std::thread::hardware_concurrency());
    }
        
    // ...

    // initialize the static shared threadpool
    threadpool& threadpool::shared_instance()
    {
        static threadpool_impl s_shared(default_thread_count());
        return s_shared;
    }

`Unfortunately, I note that a few of the unit tests fail (timeout). Specifically:

- libhttplistener_test.so : request_handler_tests:multiple_clients_multiple_requests
- libpplx_test.so : pplxtask_tests:TestTaskOperators_and_or5

It looks as the test code expects to have more threads at its disposal.

LeonineKing1199:
Well, there is something to be said for using more threads than there are physical processors.

As task-based as PPL is, each time a thread blocks on something, it's free to be rescheduled. By using context switches at the kernel level, you can avoid a lot of callback management at the software level.

I'm not sure if this is the case for the cpprestsdk but it is something to be said for the threaded model.

zerodefect:
It swings both ways. Context switches are not free. During a context switch, no useful work is done.  

My point comes from someone who is trying to maximise CPU throughput for a given piece of software in the field; minimising thread context switches is imperative.

mobileben:
I concur that there should be a means to configure the threadpool. 40 is pretty arbitrary and a hardcoded value too.

[...]

PR #441, opened 2017-05-18
Added means to change num threads in threadpool #428

mobileben:
Unfortunately I'm running on iOS, so this was forced to take some changes I needed to run in iOS. Really only threadpool.h and threadpool.cpp are the needed files. 

The other changes are to allow me to build and test.

I already submitted a PR for these iOS changes.


mobileben:
Closing since I opened a new PR based on master with no conflicts

PR #442, opened 2017-05-22
test_runner needs to be linked with libdl

madebr:
Hey,

Building cpprestsdk on Fedora, I had this issue:  Microsoft/cpprestdsk#247

This can be fixed by linking the test_runner to libdl.


ras0219-msft:
This should be fixed in the latest master, thanks for the PR!

Issue #446, opened 2017-05-26
Not able to use the wildchar "+" with http_listener


ashwinravianandan:
Hello,

I am not able to use "+" wildchar in the URI fed to the http_listener.  When I use the '*' wildchar, localhost is resolved but directly specifying the IP or 127.0.0.1 doesn't work.  The exception message I receive is:
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<
  what():  resolve: Host not found (authoritative)

Environment: linux amd64.  Library version 2.9.1 ( built based on git tag )

ashwinravianandan:
I have submitted a pull request that resolves this issue:
https://github.com/Microsoft/cpprestsdk/pull/447

ras0219-msft:
Thanks for the PR!

Issue #380, opened 2017-03-07
 HTTP listener crashes on Windows when listening on 0.0.0.0


marcofiocco:

marcofiocco:
Looks like if I listen to http://*:8000 it works, instead of http:/0.0.0.0:8000

catknight:
try it in CMD:  netsh http add urlacl url=http://*:8080/ user=BUILTIN\Users listen=yes

mahdisn76:
I have the same problem. For me these two points work:
1- you have to set http_listener url as : "http://*:port".
2- run your code as admin

chaseYLC:
"http://*" => this is loopback of ipv6 not default route. "0.0.0.0" is not ADDR_ANY but LOOPBACK. it's different.

If you want is "http:/0.0.0.0:8000", listen to "http://0:8000" instead of "http://*:8000"

PR #447, opened 2017-05-26
bugfix #446: fixed issue with wildchar handling on linux.

ashwinravianandan:


garethsb-sony:
Hi @ras0219-msft @ashwinravianandan 

What is the rationale for adding support for "+" but not "*"?
Is the difference between a "strong wildcard" and a "weak wildcard" relevant here?
Cf. base_uri::is_host_wildcard() reports true for both... (https://github.com/Microsoft/cpprestsdk/blob/master/Release/include/cpprest/base_uri.h#L350)

(Related question... Should "0:0:0:0" be mapped to a wildcard on Windows, or the wildcard approach be documented as the cross-platform solution? That has come up several times before, e.g. #380.)

PR #449, opened 2017-05-27
Supply missing WindowsTargetPlatformVersion setting...

eburkitt:
...that VS2017 insists be there. Add some NuGet droppings to .gitignore

Without this element in vcxproj an attempt to view project properties in VS2017 complains with a dialog that says "Value cannot be null.Parameter name: key".


ras0219-msft:
Thanks!

PR #454, opened 2017-05-29
Suppress Intel C++ Compiler 17.0 warning

gladkovdmitry17:
Intel C++ Compiler 17.0 concerned about missing return statement at the end of function:
`cpprest\json.h(1398): warning #1011: missing return statement at end of non-void function "web::json::number::operator=="`

buf fix of #455

### DIFF
--- a/Release/include/cpprest/json.h
+++ b/Release/include/cpprest/json.h
@@ -1383,6 +1383,7 @@ public:
                 return m_value == other.m_value;
             }
             __assume(0);
+            return false;
         }
 
     private:


ras0219-msft:
Thanks!

reneme:
I am getting compiler warnings (that are treated as errors) on Visual Studio due to this, though...

pcarik:
> I am getting compiler warnings (that are treated as errors) on Visual Studio due to this, though...

Me too, so I've had to rem out the return statement for now to get it to compile in Visual Studio

PR #456, opened 2017-05-30
fix syntax error in cmake file

EdgarWahn:
missing closing > for target_link_libraries and target_include_directories


ras0219-msft:
Thanks for the fix!

PR #460, opened 2017-06-02
implement subscript operator for const json::value

apopiak:
Note: makes m_value and m_kind mutable because there is some house keeping in those methods

### DIFF
--- a/Release/include/cpprest/json.h
+++ b/Release/include/cpprest/json.h
@@ -609,7 +609,14 @@ public:
         /// </summary>
         /// <param name="key">The name of the field</param>
         /// <returns>A reference to the value kept in the field.</returns>
-        _ASYNCRTIMP value & operator [] (const utility::string_t &key);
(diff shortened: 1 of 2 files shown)


apopiak:
signed the CLA

apopiak:
Would be great to get feedback on this, as I'm not sure it's ok style-wise (`mutable` often is sort of a code smell ;-) )

ras0219-msft:
Hi @apopiak, thanks for the PR and sorry for the delay!

You're definitely right that `mutable` is a code smell!

I think we'd prefer to mirror the C++ standard library here and only provide non-const `operator[]`: http://en.cppreference.com/w/cpp/container/map/operator_at. The trouble is that this is _fundamentally a mutating operation_. The json objects `null` and `{"x":""}` are certainly different objects, but this operator would convert the first to the second via:
```cpp
void my_innocent_function(const value& v)
{
    if (v["x"] == "things") do_things();
}
```

PR #465, opened 2017-06-07
fix build on linux use clang

fcharlie:
When use clang build on cpprestsdk on Linux, The original CMakeLists will automatically add the parameter `-stdlib=c++`  if boost is not built using `libc++`, then the entire project will fail to build. 

For almost all Linux distributions,When the boost that is installed through the Package Manager is using `GNU libstdc++` and **not** `clang libc++`. 

This PR will remove `-stdlib=libc++` on linux.

Thanks all.


fcharlie:
signed the CLA

ras0219-msft:
Thanks, looks good.

PR #476, opened 2017-06-16
Added NuGet.Config.

JunielKatarn:
Building might fail in systems with non-default (overriden) NuGet configuration.

This change ensures packages are fetched to the workspace's `packages` folder.


ras0219-msft:
Thanks!

PR #487, opened 2017-06-27
Fixed test_runner build when BUILD_SHARED_LIBS=OFF

kiyolee:
test_running always needs linking to libdl.
Found the problem when building static library on RedHat 7.


ras0219-msft:
Thanks!

PR #488, opened 2017-06-27
Ignore a bunch of temporary files from cmake/ninja.

kiyolee:
Ignore a bunch of temporary files from cmake/ninja.


ras0219-msft:
Thanks!

We would highly recommend that you use a separate build directory as well.

kiyolee:
Well, I did try to use a separate build directory, but cmake still create files inside the source tree.
Could be just another bug then.

Sent from my Windows 10 phone

kiyolee:
Just figured cmake works differently for these 2 differenct commands:
(a) cmake {dir}/CMakeLists.txt
(b) cmake {dir}
I accidentally (kind of) used the first form (that ended up not generating cmake files in its own build directory) and so lead to this pull request.

PR #494, opened 2017-07-07
Improve http_proxy handling in BingRequest sample

vadz:
This PR consists of 2 somewhat independent parts (I decided it would be more convenient to have them in the same PR, but please let me know if you want me to split it in 2): first, and most importantly, the first commit fixes the sample behaviour by default as currently it just fails to open connection if `http_proxy` is not defined, so it's broken out of the box. This is especially annoying because there is no exception handling in the sample (which is IMO far from ideal, but I didn't change this because this is, perhaps, intentional to keep things simple?), so it simply crashes without any explanation as to why does it happen.

Second, the 2nd and 3rd commits allow to use `http_proxy=auto` with this sample as it's convenient for testing.


ras0219-msft:
This seems good to me, thanks for the PR!

PR #495, opened 2017-07-08
Fix memory leak of WINHTTP_PROXY_INFO fields

vadz:
Ensure that the strings inside this struct are always freed, which wasn't the
case before.

----

According to [the documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/aa384097.aspx), the fields of this struct must be freed, so do it.


ras0219-msft:
LGTM, thanks for the fix!

Issue #182, opened 2016-06-20
web_proxy does not work with .pac files


christopher-lea-eu:
I'm trying to make my client application proxy aware, and using WinHttpGetIEProxyConfigForCurrentUser to get the proxy configuration before creating my HttpClient.
In many of my users cases this means lpszAutoConfigUrl = "http://something.com/proxy.pac".

However, it seems Casablanca does not handle this how I would expect. The .pac file is not parsed by casablanca and no correct proxy is subsequently set, as far as I can tell. A quick look through http_client_winhttp seems to suggest I can only use a direct proxy or WPAD auto discovery via DNS/DHCP?

How can I configure Casablanca to correctly use a .pac proxy? 

Thanks,
Chris

PR #496, opened 2017-07-08
Use IE proxy by default

vadz:
This is, IMO, a pretty important, even if simple, change as it makes applications C++ REST SDK to work out of the box on many corporate networks where they currently don't work by default and, if a fixed PAC URL is used, can't be made to work at all (see #182). With this change, they should work correctly in all cases, provided IE Internet access settings are configured correctly which is, IMO, always the case (at least if Internet access is allowed at all).

Please notice that this PR includes the commit already submitted as #495 as it reuses a struct declared there, so if you apply this PR entirely, you can close #495.


xqp:
ah great.

i have a similar pr for this (https://github.com/Microsoft/cpprestsdk/pull/500).
However, I like yours very well. Maybe you can check if the os version is greater 8.1 and then you can use the WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY flag to discover the proxy automatically.

👍

vadz:
Thanks, I didn't know about `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` and it indeed looks like the right way to do it under 8.1+. However I'm not sure if this flag does anything different from what my patch does manually and as I need this to work under Windows 7 too anyhow, I have to keep this code. So the question becomes whether it's still worth using the flag and make the code more complicated or if we can just use the same code under all Windows versions?

xqp:
Thanks for your answer :)

Unfortunately you call 'WinHttpOpen' with a depricated flag on Win8.1 and greater.
See the table from dwAccessType : 
WINHTTP_ACCESS_TYPE_DEFAULT_PROXY
https://msdn.microsoft.com/library/windows/desktop/aa384098(v=vs.85).aspx

vadz:
Well, the fact that it's deprecated doesn't really mean anything. It definitely still works (I've tested under Windows 10) and it's very unlikely to stop working in the observable future.

Of course, in (very) long term we should just use `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY`, but if I'm correct in that it doesn't do anything more than my manual code (this is the important question, really), then I'd prefer to switch to it once we can drop support for Windows < 8.1 entirely and _replace_ the code in my patch with it, instead of having _both_ this code and the flag.

reneme:
Thanks for your pull request! We (@xqp and me) tried the patch in our target environment and it worked fine finding the right proxy configuration!

There was issue though, that is related to WinHTTP:
When using HTTPS, WinHTTP needs to send an HTTP `CONNECT` request to the proxy. This message is not encrypted and readable by the proxy. Our target environment uses the `User-Agent:` header in this HTTP message to decide if traffic should go out or not. Unfortunately WinHTTP doesn't send this header by default and we didn't find a way to set it.

Apparently setting the user agent in the `WinHttpOpen()` call [used to work but doesn't anymore](https://stackoverflow.com/questions/2396554/send-user-agent-through-connect-and-post-with-winhttp).

Maybe you happen to know what could be a workaround for this?

vadz:
@reneme Sorry, I don't know anything about this, I didn't test with a proxy performing User-Agent checks (why would anybody do that!?) and if the Microsoft person at the linked answer says that there is no workaround, I'm inclined to believe him... Perhaps you can use ASIO backend under Windows instead of WinHTTP one? Of course, you wouldn't be able to automatically use IE proxy with it then.

ras0219-msft:
Thanks for the PR! I've merged this with #500, so the general proxy detection logic should be used both when the client is set to "Default" as well as "Auto Configure".

PR #497, opened 2017-07-08
Fix unreachable statement warning from MSVC after ICC warning fix

vadz:
Adding a return statement after __assume(0) resulted in "warning C4702:
unreachable code" when building with MSVC, which broke the build as warnings
are fatal due to the use of /WX options.

Fix this by using this return statement only with the Intel compiler, which
apparently needs it (see e3f81c436f62c46899037bc8eac4f64de3a12a15).


ras0219-msft:
Thanks!

PR #498, opened 2017-07-09
Avoid spurious errors when handling HTTP responses without body

vadz:
There is no message body for 1xx, 203 and 304 HTTP responses (see
section 3.3.3/1 of RFC 7230), yet we tried to read it nevertheless when
using ASIO backend, resulting in "Failed to read response body"
exception being thrown.

Fix this by explicitly skipping reading the body for these status codes.

----

This is (or should be, IMO) relatively high priority as it fixes a bug which currently breaks any application receiving 304 responses (which are quite common, of course) under Linux.

### DIFF
--- a/Release/src/http/client/http_client_asio.cpp
+++ b/Release/src/http/client/http_client_asio.cpp
@@ -1229,9 +1229,17 @@ private:
             }
         }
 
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thank you for including the RFC reference, it makes this much easier to review!

Issue #27, opened 2015-12-01
outside_tests hit servers with behavior that depends on country of origin


Alexander-Shukaev:
Built from the latest revision (78f8554) in the `master` branch:

```
Starting test case outside_tests:outside_cnn_dot_com...
../tests/functional/http/client/outside_tests.cpp:56: error: Failure in outside_cnn_dot_com: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
../tests/functional/http/client/outside_tests.cpp:61: error: Failure in outside_cnn_dot_com: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
Test case outside_tests:outside_cnn_dot_com FAILED
Starting test case outside_tests:outside_google_dot_com...
../tests/functional/http/client/outside_tests.cpp:73: error: Failure in outside_google_dot_com: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
../tests/functional/http/client/outside_tests.cpp:73: error: Failure in outside_google_dot_com: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
Test case outside_tests:outside_google_dot_com FAILED
Starting test case outside_tests:multiple_https_requests...
../tests/functional/http/client/outside_tests.cpp:87: error: Failure in multiple_https_requests: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
../tests/functional/http/client/outside_tests.cpp:87: error: Failure in multiple_https_requests: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
../tests/functional/http/client/outside_tests.cpp:87: error: Failure in multiple_https_requests: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
../tests/functional/http/client/outside_tests.cpp:87: error: Failure in multiple_https_requests: CHECK_EQUAL(status_codes::OK, response.status_code()) where status_codes::OK=200 and response.status_code()=302
 FAILED
[...]

PR #499, opened 2017-07-11
Fix cnn outside test

c72578:
- http://www.cnn.com redirects users from countries outside of the US
  to the "http://edition.cnn.com/" drop location
- Use http://edition.cnn.com, which does not redirect - even in the US
- Fix https://github.com/Microsoft/cpprestsdk/issues/27


ras0219-msft:
Thanks!

PR #500, opened 2017-07-12
Use system proxy

xqp:
we need a way to discover the proxy settings from os.

if the os is greater or equal to Win 8.1 we use the Automatic proxy discovery flag from windows.
Otherwise we check whether the proxy is set in the IE configuration.


ras0219-msft:
Thanks for the PR! I added an RAII wrapper to clean up the memory allocated during the call.

PR #501, opened 2017-07-12
Fix google outside tests

c72578:
- http://www.google.com redirects country specifically
- Use http://code.google.com, which does not redirect (so far)

Remark: these tests are dependent on external web sites. As soon as Google adds
redirection or changes the contents, tests will fail again.


ras0219-msft:
Thanks!

Issue #485, opened 2017-06-26
asyncrt_utils.h: xlocale.h has been renamed upstream


c72578:
Currently the file "Release/include/cpprest/asyncrt_utils.h" includes xlocale.h
Recently, xlocale.h has been renamed upstream in glibc. See:

Rename xlocale.h to bits/types/__locale_t.h
https://sourceware.org/git/?p=glibc.git;a=commit;h=f0be25b6336db7492e47d2e8e72eb8af53b5506d
https://sourceware.org/git/?p=glibc.git;a=commit;h=af85385f311c574590381f7897461643f8a2fe6f

cpprestsdk should care about this include, ideally check for it.

Otherwise there will be the following build error:
```
/builddir/build/BUILD/cpprestsdk-2.9.1/Release/include/cpprest/asyncrt_utils.h:33:10:
fatal error: xlocale.h: No such file or directory #include <xlocale.h>
```

PR #502, opened 2017-07-13
Fix build error with glibc 2.26, xlocale.h

c72578:
- Do not include xlocale.h on systems, where __GLIBC__ is defined
  xlocale.h has been removed from glibc 2.26
  The include of locale.h in asyncrt_utils.h is sufficient
  Further details:
  https://sourceware.org/git/?p=glibc.git;a=commit;h=f0be25b
- Fixes https://github.com/Microsoft/cpprestsdk/issues/485


ras0219-msft:
LGTM, thanks for the PR!

PR #504, opened 2017-07-14
Make web::uri_builder::to_string() and ::to_uri() const

reneme:
There is no reason for those to be non-const. We would like to call them inside a `const` context. :)


ras0219-msft:
Thanks, agreed :).

Issue #235, opened 2016-09-09
Get client IP address on http_listener 


jestebancds:
I can't find a way to obtain the IP address of incomming transaction

orudge:
This would be a very useful feature. I think I've found what would be needed to implement support for this for the Windows/HTTPAPI implementation. If I'm able to come up with a decent solution for this I'll post a comment here, if I have time to look at it...

jestebancds:
I looking for linux enviroment, but if you come up with something for Windows I could adapted to linux implementation

adenisevich:
any news ?

Issue #445, opened 2017-05-26
Is there any solution for getting client's IP address in http server side?


Jaehak0123:
Hi, I'm currently programming http server side with cpprest.

And, I've catched that I can't get client's IP address which was connected.
Is there any solution for getting client's IP address?

PR #507, opened 2017-07-17
Add support for retrieving remote address in HTTP listener

orudge:
This PR fixes issues #235 and #445 by providing a mechanism to allow the client's IP address to be retrieved in the HTTP listener. I've also added in a basic unit test - more complex tests are tricky due to the unknown testing environment unfortunately.

Tested on Windows 7 and on Ubuntu.

### DIFF
--- a/Release/include/cpprest/http_msg.h
+++ b/Release/include/cpprest/http_msg.h
@@ -715,6 +715,8 @@ public:
 
     _ASYNCRTIMP void set_request_uri(const uri&);
 
+    const utility::string_t& remote_address() const { return m_remote_address; }
+
     const pplx::cancellation_token &cancellation_token() const { return m_cancellationToken; }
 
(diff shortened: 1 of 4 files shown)


orudge:
As a note, I've made a further revision which uses the RtlIpv6AddressToStringExW function (instead of InetNtopW), to ensure that the IPv6 scope ID is included. This changeset is based off a v2.9.1 branch but I can resubmit the pull request if you think this would be a better solution.

https://github.com/orudge/cpprestsdk/commit/c472852ea0ad0e9b36eaead50abad730cb55e8f8

ras0219-msft:
Thanks for the PR and sorry for the delay!

It looks like `RtlIpv6...` is primarily for driver usage, so I'd prefer to keep using `InetNtopW` if it still functions acceptably for #235 and #445.

PR #508, opened 2017-07-17
Adding 308 "Permanent Redirect" to http_constants.dat

deeringc:
As per https://tools.ietf.org/html/rfc7538#section-3


ras0219-msft:
Thanks!

PR #509, opened 2017-07-20
fix lambda capture 'this' is not used, clang 5.0 on ubuntu 16.04

fcharlie:
By default on clang 5.0, if lambda capture 'this' is not used, clang report error, this pr fixed

### DIFF
--- a/Release/src/uri/uri.cpp
+++ b/Release/src/uri/uri.cpp
@@ -22,12 +22,12 @@ utility::string_t uri_components::join()
     // canonicalize components first
 
     // convert scheme to lowercase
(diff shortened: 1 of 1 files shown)


ras0219-msft:
This changes the behavior of the lambda in a suboptimal way -- we do not need to capture `this` for the behavior we need.

What is the exact text of the error? If there's no way to work around it in code besides adding a capture, we should instead suppress it.

fcharlie:
My clang revision is svn: `308582` , branch `release_50`

ras0219-msft:
I'm blind, thank you for the PR!

PR #516, opened 2017-08-03
enable certificate revocation check with winhttp

xqp:
-enable the crl check in cpprestsdk (use the right handle type)
-delete unused and not working code (the code use the session handle but need the request handle)


ras0219-msft:
Thanks for the PR!

ulrichb:
What about Linux (when cpprestsdk uses OpenSSL for TLS)? By observation it seems that there also happen no certificate revocation checks 🤔

Issue #517, opened 2017-08-08
win32_encryption throws exception when string is empty


arno5270:
When on Windows you try to create credentials with an empty password, an exception is thrown. The password is encrypted using the win32_encryption class, which calls CryptProtectMemory, which does not handle zero length buffers. Instead, it will return error code 87 (incorrect parameter).

PR #521, opened 2017-08-16
Fix/win32 encryption throws by empty string

xqp:
- do a early return if the string is empty
- i've add some tests for this class

[Link to the Issue](https://github.com/Microsoft/cpprestsdk/issues/517)


ras0219-msft:
LGTM, thanks for the PR!

PR #527, opened 2017-08-23
Fixed proxy credentials for http_client_asio

ChristophAlbert:
On Linux, the HTTP client uses the wrong credentials to create the proxy_authorization header.


ras0219-msft:
Thanks for the PR!

Issue #443, opened 2017-05-23
building static from git on debian linux fails


EdgarWahn:
Target:
get cpprestsdk via git and directly compile static version of lib

Prerequisites:
installed all dependecies as described in https://github.com/Microsoft/cpprestsdk/wiki/How-to-build-for-Linux
Make makefile through cmake: cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=0

Result:
Make breaks when building the included test / sample applications.
Linking CXX executable ../../../Binaries/test_runner
/usr/bin/ld: CMakeFiles/test_runner.dir/test_module_loader.cpp.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status

vhwang:
I am running into the exact same problem with Ubuntu 16.04.02 and Fedora when trying to build a static version of the lib.

mobileben:
@EdgarWahn I believe the issue is the it is missing libdl. I've created a PR for this. Note you can also turn off tests and samples via ```-DBUILD_SAMPLES=OFF -DBUILD_TESTS=OFF``` (which is how I usually build).

ras0219-msft:
This issue was fixed by 265421751222207a5f32dfc79ef1b4fc94e10dcc. Thank you for reporting this problem!

PR #541, opened 2017-09-12
add in link for CMAKE_DL_LIBS for linux linking #443

mobileben:
Issue with #443 is related to the samples not including libdl. There are a few ways to do this, and this option avoided having to modify the CMakeLists.txt file for the samples.


ras0219-msft:
This should be completely taken care of by e5350bac8904286c5b and 2654217. Thank you for the PR regardless!

PR #543, opened 2017-09-18
Adding "429 - Too Many Requests" to the http constants

deeringc:
As per https://tools.ietf.org/html/rfc6585


ras0219-msft:
Thanks!

PR #550, opened 2017-10-01
Add all remaining missing HTTP status codes

deeringc:
I had been adding PRs for these as I encountered missing ones in my usage. I encountered another so I figured I would just go ahead and add all the remaining ones in one go.

I left out 418 - "I'm a teapot", although I was tempted! ;)


ras0219-msft:
LGTM, thanks!

PR #551, opened 2017-10-01
Adding has_<type>_field utility to json value

deeringc:
In our code base we have a repeating pattern of accessing json objects while interpreting server responses.

It looks something like as follows:

`
if(val.has_field(U("key")) && val.at(U("key")).is_string())
{
   something =  val.at(U("key")).as_string()
}
`

Both of these checks are required or a json exception will be thrown at runtime. This is quite verbose and problematic as sometimes developers on our team forget to check both of these. I'm proposing adding a new check for each type supported in json that combines these two checks into one.

With this change the above code becomes:
`
if(val.has_string_field(U("key")))
{
   something =  val.at(U("key")).as_string()
}
`

### DIFF
--- a/Release/include/cpprest/json.h
+++ b/Release/include/cpprest/json.h
@@ -544,6 +544,55 @@ public:
         /// <returns>True if the field exists, false otherwise.</returns>
         bool has_field(const utility::string_t &key) const;
 
+		/// <summary>
+		/// Tests for the presence of a number field
+		/// </summary>
+		/// <param name="key">The name of the field</param>
+		/// <returns>True if the field exists, false otherwise.</returns>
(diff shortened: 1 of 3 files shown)


ras0219-msft:
These seem quite reasonable to me as is, though they leave a bit of performance on the table (in the pattern you have, it makes 3 separate lookups).

I think to fix this would really require `std::optional`, however, so I think these APIs could live alongside some future `.maybe_get_string_field()` just fine.

deeringc:
Agreed, once CppRestSDK uses c++17 then std::optional is the way to go! Actually, I can make this more efficient so that it does one less lookup. I'll follow up with another PR. Thanks for the feedback.

PR #562, opened 2017-10-17
Fix building with LibreSSL

aduskett:
LibreSSL is backwards compatible with OpenSSL 1.0, however it reports
it's version number as 2.0, which then causes build failures when using
OPENSSL_VERSION_NUMBER.

Adding a check for LIBRESSL_VERSION_NUMBER when checking if the OpenSSL
version number is < 0x10100000L fixes the issue.


ras0219-msft:
LGTM, thanks for the PR!

PR #565, opened 2017-10-19
Add support for retrieving HTTP version of a request in HTTP listener

### LINKED ISSUE #545: Change to remote address introduced bug that crashes the server if client disconnects
The following change to `http_server_asio.cpp` introduced an issue:

https://github.com/Microsoft/cpprestsdk/commit/2dcf4fd187d2ca6ce32a9b1ad9b01bcdb8fb08d5#diff-d8600089329e519f56170a2d16d7102c

When the client connects and immediately disconnects, `remote_endpoint().address()` will throw an exception (ENOTCONN):

    terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
      what():  remote_endpoint: Transport endpoint is not connected

This crashes the server.

garethsb-sony:
This PR follows on from #507 by providing a mechanism to allow the HTTP version of the incoming message to be retrieved in the HTTP listener. (One motivation is to allow construction of access logs in Common Log Format.) A basic unit test is included.

It also changes http_request::get_remote_address() introduced by #507 to http_request::remote_address() for consistency with pre-existing member functions of this class.

It also includes a commit that fixes #545 which has been reported against #507.

### DIFF
--- a/Release/include/cpprest/http_msg.h
+++ b/Release/include/cpprest/http_msg.h
@@ -921,6 +921,9 @@ public:
     /// <returns>The remote IP address.</returns>
     const utility::string_t& remote_address() const { return _m_impl->remote_address(); }
 
+    CASABLANCA_DEPRECATED("Use `remote_address()` instead.")
+    const utility::string_t& get_remote_address() const { return _m_impl->remote_address(); }
+
     /// <summary>
     /// Extract the body of the request message as a string value, checking that the content type is a MIME text type.
     /// A body can only be extracted once because in some cases an optimization is made where the data is 'moved' out.
--- a/Release/include/cpprest/http_msg.h
+++ b/Release/include/cpprest/http_msg.h
@@ -731,7 +731,7 @@ public:
 
     _ASYNCRTIMP void set_request_uri(const uri&);
 
-    const http_version& http_version() const { return m_http_version; }
+    const http::http_version& http_version() const { return m_http_version; }
 
     const utility::string_t& remote_address() const { return m_remote_address; }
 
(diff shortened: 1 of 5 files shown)


ras0219-msft:
Thanks!

I changed `http_version` to be a full `struct` instead of just using a `pair`, for bonus type safety. I also restored the older `get_remote_address()` to avoid breaking source compatibility (with a deprecation warning).

garethsb-sony:
Thanks for working on these, @ras0219-msft. I think there is one problem with the struct http_version commit. operator>> with uint8_t doesn't do what we want, it reads a character. If I'm right, we need to restore the temporary unsigned ints, around here: https://github.com/Microsoft/cpprestsdk/pull/565/commits/1d358474c02e2c4c3e2043b97e04d0777be2fb90#diff-d8600089329e519f56170a2d16d7102cR657.

ras0219-msft:
You are absolutely right. I'll address this tomorrow or I'd be happy to merge a PR.

garethsb-sony:
Perhaps it's worth pulling the parsing code out into a unit testable function?
E.g. add ``from_string()`` and ``to_string()`` to`` http_version``?
(Or ``op>>`` and ``op<<``?)

garethsb-sony:
Ah, the existing unit test is failing when cpprestsdk is configured with ```CPPREST_HTTP_LISTENER_IMPL``` as ```asio```, so that's something!

```
...\tests\functional\http\listener\request_handler_tests.cpp(466): error : Failure in http_version:
CHECK_EQUAL(true, httpVersion == http_versions::HTTP_1_1) where true=1 and httpVersion == http_versions::HTTP_1_1=0
```

PR #566, opened 2017-10-19
Fix an issue where requests sent via authenticated proxies could not be redirected

deeringc:
There is a scenario where a client connects to a webserver via an authenticated proxy (requiring user supplied credentials) and is then redirected (eg via a HTTP 302) to another webserver. 

When the request is redirected to another webserver then WinHTTP will need to traverse the authenticated proxy again. However, the current code doesnt handle this well. 

It currently sets a bool flag to indicate that the proxy creds have already been supplied once (during the first request) and so should not be applied again (so as to avoid a 407 loop with the proxy in the case where the credentials are wrong). 

The logic needs to be more nuanced however, to allow the supplying of the credentials to WinHTTP in the case of a redirect, to allow the 2nd connection to traverse the authenticated proxy.

### DIFF
--- a/Release/src/http/client/http_client_winhttp.cpp
+++ b/Release/src/http/client/http_client_winhttp.cpp
@@ -1038,6 +1038,25 @@ private:
         }
     }
 
+    static utility::string_t get_request_url(HINTERNET hRequestHandle)
+    {
+        DWORD urlSize{ 0 };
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Yep, I agree with the logic. Thanks for the PR and the explanation! Sorry for the delay!

PR #569, opened 2017-10-23
fix compiling with non-glib Linux c libraries

aduskett:
This commit fixes building against uclibc and musl c libraries.

- musl requires sys/time.h
- musl and uclibc don't define __GLIBC__, and don't include xlocal.h.

Instead of adding more define checks, have cmake check for the header and
set the define "HAVE_XLOCALE_H", then check for that instead.

### DIFF
--- a/Release/CMakeLists.txt
+++ b/Release/CMakeLists.txt
@@ -42,6 +42,7 @@ include(cmake/cpprest_find_boost.cmake)
 include(cmake/cpprest_find_zlib.cmake)
 include(cmake/cpprest_find_openssl.cmake)
 include(cmake/cpprest_find_websocketpp.cmake)
+include(CheckIncludeFiles)
 
 find_package(Threads REQUIRED)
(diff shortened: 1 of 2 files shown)


aduskett:
Signed the CLA.  Thanks!

c72578:
I just tested this PR and it is working nicely here on Fedora systems with and without xlocale.h:
Fedora 26, glibc 2.25:
-- Looking for include file xlocale.h
-- Looking for include file xlocale.h - found

Fedora 27, glibc 2.26
-- Looking for include file xlocale.h
-- Looking for include file xlocale.h - not found

OK

aduskett:
Thanks for testing it!  I am trying to get the library into BuildRoot and this patch is necessary to do so.  I always like to upstream my patches whenever it's relevant!

aduskett:
Bump?

ras0219-msft:
Thanks! I moved the `<sys/time.h>` include to not break windows. Please submit a followup PR if needed to adjust it further :)

PR #585, opened 2017-11-03
Resolve gcc "warning: comparison between signed and unsigned integer expressions [-Wsign-compare]"

garethsb-sony:
Some functions in cpprest/details/fileio.h return (unsigned) size_t but are documented with e.g. <returns>0 if the read request is still outstanding, -1 if the request failed, otherwise the size of the data read into the buffer</returns>.


ras0219-msft:
Thanks for the fix!

Issue #358, opened 2017-02-14
WebSocket server implementation?


garethsb-ghost:
I had been hoping to find a websocket server implementation to use alongside http_listener. Have I missed it, is it in the roadmap, or do you have any pointers on doing it myself (I'm currently looking at mashing something up wrapping websocketpp::server)?

Thanks!

ampyx:
+1

gdelab:
+1

garethsb-sony:
Please see if the interface at https://github.com/sony/nmos-cpp/blob/master/Development/cpprest/ws_listener.h and implementation at https://github.com/sony/nmos-cpp/blob/master/Development/cpprest/ws_listener_implementation.cpp is of interest. It may not be complete enough for cpprestsdk yet, but feedback is welcome.

sinall:
@garethsb-sony do you plan to send a pull request to cpprestsdk? It will help a lot of people.

garethsb-sony:
Yes. Sorry, it will take us a little while to prepare to do it, but any feedback is welcome.

hdiethelm:
+1 for this!
I implemented the function set_message_handler(), so the server can also handle incoming messages:
[ws.diff.txt](https://github.com/Microsoft/cpprestsdk/files/1732548/ws.diff.txt)

nathanprins:
@garethsb-sony are you still intending on sending a pr? It has been some time since your previous comment.

garethsb-sony:
@nathanprins There seems little interest from the Microsoft maintainers in adding new functionality, and I think it unlikely a PR would be accepted without a test suite in the C++ REST SDK style. However, my implementation is in active use on several projects and is now reasonably feature complete. See interface at https://github.com/sony/nmos-cpp/blob/master/Development/cpprest/ws_listener.h and implementation at https://github.com/sony/nmos-cpp/blob/master/Development/cpprest/ws_listener_impl.cpp.

nathanprins:
Alright, I'll take a closer look at your repo. Thanks!

garethsb-sony:
@nathanprins - recently updated to enable richer user validation on connection, e.g. to check "Authorization" header when doing Secure WebSocket.

jbbcst:
Hi all,

[...]

PR #598, opened 2017-11-07
Add cmake/CMakeHelpers.cmake, previously omitted due to websocketpp's over-eager top-level .gitignore

garethsb-sony:
Seems obvious how this happened!

Source file is here:
https://github.com/zaphoyd/websocketpp/blob/0.5.1/cmake/CMakeHelpers.cmake

### DIFF
--- a/Release/libs/websocketpp/cmake/CMakeHelpers.cmake
+++ b/Release/libs/websocketpp/cmake/CMakeHelpers.cmake
@@ -1,0 +1,78 @@
+
+# Print build configuration
+macro (print_used_build_config)
+    message ("\n=========== Used Build Configuration =============\n")
+    message (STATUS "ENABLE_CPP11        = " ${ENABLE_CPP11})
+    message (STATUS "BUILD_EXAMPLES      = " ${BUILD_EXAMPLES})
+    message (STATUS "BUILD_TESTS         = " ${BUILD_TESTS})
+    message ("")
+    message (STATUS "WEBSOCKETPP_ROOT    = " ${WEBSOCKETPP_ROOT})
+    message (STATUS "WEBSOCKETPP_BIN     = " ${WEBSOCKETPP_BIN})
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thanks for this PR, however I don't understand how it affects cpprestsdk at all.

We don't use the CMakeLists.txt for websocketpp, so we don't need the helpers. If you'd like to use websocketpp, I highly recommend you pull the latest version directly and configure us to use it instead of this embedded copy.

garethsb-sony:
That's a fair point. Here's some background.

I have a reasonable sketch of an implementation of a websocket server for cpprestsdk (which I'm currently calling ``web::websockets::experimental::listener::websocket_listener`` in cpprest/ws_listener.h), utilising WebSocket++ just as the existing ``websocket_client`` does.

I wanted to make sure this functioned with the embedded websocketpp on Windows and Linux before preparing a PR - if such a PR would be considered? (Discussion in #358.)

The embedded websocketpp has a few small differences from original v0.5.1, so that seemed like a good idea, and since I'm using CMake to build my own project - and now cpprestsdk too of course! - I thought I'd just point CMake at the embedded websocketpp. It didn't work, but this single missing file is the only reason why.

garethsb-sony:
That's a fair point. Here's some background.

I have a reasonable sketch of an implementation of a websocket server for cpprestsdk (which I'm currently calling ``web::websockets::experimental::listener::websocket_listener`` in cpprest/ws_listener.h), utilising WebSocket++ just as the existing ``websocket_client`` does.

I wanted to make sure this functioned with the embedded websocketpp on Windows and Linux before preparing a PR - if such a PR would be considered? (Discussion in #358.)

The embedded websocketpp has a few small differences from original v0.5.1, so that seemed like a good idea, and since I'm using CMake to build my own project - and now cpprestsdk too of course! - I thought I'd just point CMake at the embedded websocketpp. It didn't work, but this single missing file is the only reason why.

Issue #595, opened 2017-11-05
Linux build fails with GCC [7.2.1]  with  [-Werror=type-limits]


entd:
Library build fails on late version GCC:
`~/src/casablanca/Release/src/uri/uri.cpp: In static member function ‘static utility::string_t web::uri::decode(const string_t&)’:
~/src/casablanca/Release/src/uri/uri.cpp:676:24: error: comparison is always false due to limited range of data type [-Werror=type-limits]
         else if (*iter > CHAR_MAX || *iter < 0)
`
While it could be supressed it imply deeper problem with code. For me comparing if 8 bit char is more than CHAR_MAX seems pointless too. Is it false positive?

echotor:
Hi, I want to know this problem is solved or not.

StrangeTcy:
I'm having the same problem and am similarly interested in a solution.

codehero:
This issue occurs with GCC 5.4.0 as well

promgamer:
Same here. Trying to compile with gcc 7 for C++17.

sridherj:
Did it work with C++17?

ras0219-msft:
I've pushed a fix for this to the master branch, thanks everyone for reporting it!

PR #599, opened 2017-11-08
fix build by gcc error: -Werror=type-limits

fcharlie:
Fix https://github.com/Microsoft/cpprestsdk/issues/595

We known , `CHAR_MAX`= 127, ascii `0~127`.


fcharlie:
@ras0219-msft Please check thanks.

PR #606, opened 2017-11-14
Make it possible to set the user agent for websockets

MathyV:
Since I don't have a box to test winrt implementation I can't add/test that unfortunately


garethsb-sony:
I'm not sure, but can this be done already with ``my_config.headers().add(web::http::header_names::user_agent, my_user_agent)``?

MathyV:
I tried but websocketpp overwrites it as far as I can tell if you don't specify it explicitly.

PR #608, opened 2017-11-23
Define http_listener if CPPREST_FORCE_HTTP_LISTENER_ASIO is defined

garethsb-sony:
When cpprestsdk is built with ``CPPREST_HTTP_LISTENER_IMPL = asio`` on Windows, for testing purposes (I'm trying to debug an issue encountered on Linux), the following complete source code fails to compile:

```C++
#include "cpprest/http_listener.h"
using web::http::experimental::listener::http_listener;
```

This is a one-line fix. (It's possible that the various preprocessor conditional expressions in this header should all be made more consistent.)

### DIFF
--- a/Release/include/cpprest/http_listener.h
+++ b/Release/include/cpprest/http_listener.h
@@ -21,7 +21,7 @@
 #include <boost/asio/ssl.hpp>
 #endif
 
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Looks simple, thanks for the PR!

I agree, the preprocessor macros are probably in need of a simplification.

PR #609, opened 2017-11-23
Support for GCC 4.7 (and VS2013)

garethsb-sony:
Only a small number of changes are required to build C++ REST SDK for GCC 4.7. While this may not be an officially supported compiler, the changes in this PR are ultimately trivial, with little or no effect on currently supported versions of GCC and MSVC. Each commit contains the  workaround for one issue, explained in the commit comment.

One of the workarounds is also applied for VS2013 which has a similar compiler bug.


ras0219-msft:
These all look good; I'll be happy to merge them once you've signed the CLA!

garethsb-sony:
@ras0219-msft Hi Robert, I just received an early Christmas present, the go-ahead to sign the CLA! I would be very happy if you would be able to merge PRs #609, #611, #612 and #624,  and I would welcome any feedback on PRs #630 and #565. Thanks again for your patience.

ras0219-msft:
Thanks!

PR #618, opened 2017-12-01
Update repology badges

AMDmi3:


ras0219-msft:
Thanks for the PR! I actually already fixed them in master before noticing this :)

PR #624, opened 2017-12-07
Expose http_listener configuration for the backlog

garethsb-sony:
When a large number of concurrent connection requests need to be dealt with, configuring the backlog (the maximum length of the queue of pending connections on the port) can prevent the TCP stack refusing connections.

Tested on both Windows 8+ (see https://blogs.msdn.microsoft.com/winsdk/2015/06/01/winsocks-listen-backlog-offers-more-flexibility-in-windows-8/) and Linux with the ASIO-based http_listener.

I haven't been able to determine whether/how the HTTP Server API Version 2.0 on Windows allows this to be configured.

### DIFF
--- a/Release/include/cpprest/http_listener.h
+++ b/Release/include/cpprest/http_listener.h
@@ -46,6 +46,7 @@ public:
     /// </summary>
     http_listener_config()
         : m_timeout(utility::seconds(120))
+        , m_backlog(0)
     {}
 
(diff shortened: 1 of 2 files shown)


ras0219-msft:
Looks good to me, pending CLA

ras0219-msft:
Thanks for the PR!

Issue #577, opened 2017-10-31
Inconsistent DLL naming in CMake vs. vcxproj builds at v2.10.0?


garethsb-sony:
As was the case for v2.9, at v2.10.0, the TargetName set in the project files in the repo is like "cpprest140d_2_10". E.g. https://github.com/Microsoft/cpprestsdk/blob/master/Release/src/build/vs14/casablanca140.vcxproj#L37

However, after using CMake to generate project files instead, the TargetName in the resulting cpprest.vcxproj is simply "cpprest_2_10", i.e. without the VS version or DebugFileSuffix.

Is this intentional, or should CMake builds include the VS/debug disambiguators in the same way?

Thanks.

garethsb-sony:
The repo .vcprox files also have Win32 and x64 (and ARM) builds. I understand a CMake-generated project supports a single platform, but I wonder whether the output directory should be made to still include e.g. "x64" in the path, as before?

PR #630, opened 2017-12-15
Patch for Windows DLL naming

garethsb-sony:
Patch to get e.g. "cpprest140d_2_10.dll" as described in #577.


ras0219-msft:
Thanks for the PR!

I simplified the abi-tag calculation to make it more compatible with the default cmake way of operating (i.e. using `CMAKE_DEBUG_POSTFIX`). While this does result in a different binary name than building with our embedded `.sln`, the important capability of distinguishing debug and release binaries is still preserved.

PR #634, opened 2017-12-20
Fix uncaught exception in http_server_asio.cpp for invalid URL path

garethsb-sony:
URLs that contain invalid UTF-8 characters, either before or after percent-decoding, can result in ``std::range_error``, which the ASIO-based http_listener does not handle.

E.g. "/%f8%80%80%80%ae%f8%80%80%80%ae/%f8%80%80%80%ae%f8%80%80%80%ae/.../boot.ini".


ras0219-msft:
Looks good, thanks for the PR!

PR #636, opened 2017-12-22
Fix string use after destruct introduced in merge commit b7fa0b1

JonathanGawrych:
See the comment on L480. If the uri port is not default, and is greater than 0, then a LPCWSTR (proxy_name) is pointed to a string that is then destructed before it is used. Reverted the string declaration move, so that it's lifetime is preserved after proxy_name is used.


ras0219-msft:
Yep, this is the right fix. Thanks!

PR #638, opened 2017-12-26
fixed strand type to be able to build it with Boost 1.66

jhruby:
I've changed strand type to one that is not deprecated to be able to build the library with Boost 1.66


aduskett:
This is a good fix. Confirmed to work when cross compiling with BuildRoot as well.

Issue #643, opened 2018-01-03
ssl proxy port is hard coded to 443


sidarthagracias:
When using the http client to connect through a http proxy using ssl, I see the cpp rest sdk has the ssl port hard coded to 443, specifically this file 
Release/src/http/client/http_client_asio.cpp
   _const auto &base_uri = m_context->m_http_client->base_uri();
   const auto &host = base_uri.host();  
   request_stream << "CONNECT " << host << ":" << 443 << " HTTP/1.1" << CRLF;
   request_stream << "Host: " << host << ":" << 443 << CRLF;
   request_stream << "Proxy-Connection: Keep-Alive" << CRLF;_

Is there any reason we decided not to pull port information from the base_uri
i.e. int port = base_uri.port();

dimarusyy:
Hi, I had the same issue and updated the code (similar to whata you suggested) :
```
const auto &port = base_uri.port();
request_stream << "CONNECT " << host << ":" << (( port != 0 ) ? port : 443) << " HTTP/1.1" << CRLF;
request_stream << "Host: " << host << ":" << (( port != 0 ) ? port : 443) << CRLF;

```

ras0219-msft:
This is definitely a bug and I'd be happy to accept a PR fixing it!

dimarusyy:
I'm not in "contributors" list. Attached [diff.txt](https://github.com/Microsoft/cpprestsdk/files/1602860/diff.txt)

dimarusyy:
https://github.com/Microsoft/cpprestsdk/pull/646

PR #646, opened 2018-01-04
#643 : ssl proxy port is hard coded to 443

dimarusyy:
- use base URI port in "CONNECT" request to proxy server


ras0219-msft:
Thanks!

PR #662, opened 2018-01-16
Overhaul the build process for OpenSSL dependency on Android

jwtowner:
The root Makefile for building OpenSSL has been rewritten, merging
the behavior previously found in the external setenv-android.sh
shell script, with the following numerous changes and additions to
allow for more varied build configurations:

- Added support for building with the Clang/LLVM toolchain instead
  of GCC to the Makefile.
- Added support for more recent Android NDK releases, including
  r10e, r11c, r12b, r13b, r14b, r15c, and r16b.
- Added support for the AArch64 Android target.
- Added support for more OpenSSL releases, with patch files for
  OpenSSL 1.0.2k through 1.0.2n and preliminary support for
  OpenSSL 1.1.0g to ensure forward compatibility.
- Most of the configuration variables in relation to Android now
  follow the same naming scheme and values as CMake.
- Exposed more configuration options in the Makefile as user
  overridable variables.
- The Makefile must now be invoked once per build target, it no
  longer builds libraries for both armv7 and x86 in the same
  invocation. This simplifies the build rules in the Makefile.
- Building OpenSSL within Ubuntu on WSL now completes
  successfully so long as the Android SDK, Android NDK, and
  cpprestsdk source files are stored on an Ext4/Ext3/Ext2 filesystem.
  Before it was failing within setenv-android.sh due to it
  containing trailing carriage returns.

The default build configuration for OpenSSL remains the same as
before, with GCC 4.8 as the toolchain, Android-18 as the API level
and OpenSSL 1.0.2k as the OpenSSL release. At a later date,
configure.sh can be updated with additional command line
options to further expose the new features.

### DIFF
--- a/Build_android/configure.sh
+++ b/Build_android/configure.sh
@@ -108,8 +108,8 @@ then
     if [ ! -d "openssl" ]; then mkdir openssl; fi
     cd openssl
     cp -af "${DIR}/openssl/." .
-    make all ANDROID_NDK="${NDK_DIR}" ANDROID_ABI=armeabi-v7a
-    make all ANDROID_NDK="${NDK_DIR}" ANDROID_ABI=x86
+    make all ANDROID_NDK="${NDK_DIR}" ANDROID_ABI=armeabi-v7a OPENSSL_PREFIX=armeabi-v7a
+    make all ANDROID_NDK="${NDK_DIR}" ANDROID_ABI=x86 OPENSSL_PREFIX=x86
 )
 fi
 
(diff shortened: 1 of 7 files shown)


jwtowner:
Tested this with android-ndk-r16b to ensure forward compatibility, and it has removed the deprecated API-level specific header files and replaced them with unified headers in a common sysroot directory. I'm updating this PR to support use of the unified headers if they are detected to exist in the NDK, falling back to the API-level specific header files for older NDKs.

For more information see:

https://android.googlesource.com/platform/ndk/+/ndk-release-r16/docs/UnifiedHeaders.md

jwtowner:
One other minor change, forgot to remove the job count parameter I was using for testing purposes from the Makefile.

PR #666, opened 2018-01-17
issue 665, proposed solution

EdgarWahn:
- add \<project\>Config.cmake.in and \<project\>ConfigVersion.cmake.in to allow cmake to import the cppresksdk project by using find_package(\<name\> \<version\>)
- add keywords in CMakeLists.txt to produce \<project\>Config.cmake and \<project\>ConfigVersion.cmake from their .in templates
- add cmake policy CMP0048 to allow version number in project statement


ras0219-msft:
Thank you for the PR!

Please retarget this against master. We do already have some of what is proposed here.

EdgarWahn:
Noobish question... "retarget against master" means I check out the current master, apply my patches and submit the diff against what?! Sorry, just got lost.

ras0219-msft:
Yes, please check out master and apply your patches, then submit a PR here against the master branch. Thanks!

EdgarWahn:
@see 680, is this ok now?

PR #678, opened 2018-01-25
Fix VS2013 compilation errors

### LINKED ISSUE #674: VS2013 doesn't support defaulted move constructor/assignment operator
VS2013 doesn't fully support explicitly defaulted functions. See https://msdn.microsoft.com/en-us/library/dn457344.aspx#Anchor_1.

Commit b0634729ae3ece059f5d727e6176f9bf8b9a0e77 therefore results in several errors beginning with:

```
error C2610: 'web::details::uri_components::uri_components(web::details::uri_components &&)' :
is not a special member function which can be defaulted
```
at ...\cpprestsdk\Release\include\cpprest\base_uri.h(35)

It's frustrating but if we want to keep VS2013 support, the hand-written move constructors and assignment operators need to be restored, at least conditionally. (If it's even worth making this conditional, then according to https://github.com/boostorg/config/pull/205, which will add ``BOOST_NO_CXX11_DEFAULTED_MOVES`` to Boost.Config, a suitable condition might be ``_MSC_FULL_VER < 190023026``.)

### LINKED ISSUE #675: VS2013 compile fail in cpprest/streams.h
Commit 05e474a2d8f041f56691a4e9f7f79bd0f110809c causes compile to fail in VS2013, for example at cpprest/streams.h(895):

```
error C2899: typename cannot be used outside a template declaration
```

That's probably a bug in VS2013, but I'm pretty confident all the typenames and typedefs used in this file can be hugely simplified while regaining VS2013 compatibility. I'll submit a PR.

garethsb-sony:
Fix #674
Fix #675


garethsb-sony:
Hmm, not sure why the diff for commit ef4ab5987ca43c7216aa40bfada4f1b7392eab7e shows the whole file has changed? Maybe an EOL problem?

ras0219-msft:
Thanks!

PR #681, opened 2018-01-30
http_version::from_string argument is now std::string

garethsb-sony:
Fix to work with commit 1e4717e5aefb080bfc3edd3b9d6ca6e2546c0111. (Follow up to #677. )


ras0219-msft:
LGTM, Thanks!

Issue #664, opened 2018-01-16
Certificate revocation checking cannot be disabled.


kobykahane:
PR #516 enabled certificate revocation checking on WinHTTP by calling `WinHttpSetOption` with `WINHTTP_OPTION_ENABLE_FEATURE` and `WINHTTP_ENABLE_SSL_REVOCATION`.

However, if a client wants to opt out from this behavior, the callback invoked by `invoke_nativehandle_options` and set with `set_nativehandle_options` cannot be used. This is because subsequent calls to `WINHTTP_OPTION_ENABLE_FEATURE` are additive, so even setting the input flags to `0` does not reset the previously enabled 'enable SSL revocation' option.

Furthermore, revocation checks are performed even when `set_validate_certificates(false)` has been called.

Instead, `http_client_config` should provide an explicit way to opt out of revocation checks. If the appropriate flag is set, it should never call `WinHttpSetOption` with `WINHTTP_ENABLE_SSL_REVOCATION`.

tishion:
agree with this totally

saifbohra:
agree with this totally, I had the same issue

sinall:
I have the same problem. This is a critical issue.

tvarsis:
Getting the REV_FAILED error on thousands of machines after the ISRG root cert from Lets Encrypt was added to the trusted root store in Windows by Microsoft a couple of days ago. So this issue is critical to fix now.

tvarsis:
So I finally mange to fix the error. It was related to OCSP. I added this to the Apache vhost:

SSLUseStapling on
SSLStaplingReturnResponderErrors off
SSLStaplingResponderTimeout 5

As well as this to the ssl.conf file:
SSLStaplingCache "shmcb:/opt/bitnami/apache2/logs/ssl_stapling(128000)"

And that got rid of the error. Did not see this error before ISRG was in the Windows root store, but at least this solves it if anybody else runs into similar issues.

PR #687, opened 2018-02-05
Disable cert revocation check if cert validation is disabled

ZekeSnider:
This addresses an issue described in #664. It makes it so that certificate revocation is not checked if certificate validation is disabled on http_client_config. Previously there was no way to disable this functionality.


kobykahane:
@ZekeSnider In my opinion, a complete solution to issue #664 would also allow disabling certificate revocation checking without completely disabling certificate validation as well.

PR #693, opened 2018-02-07
Don't use the ``U()`` preprocessor macro in public headers

### LINKED ISSUE #692: Compilation errors when using cpprestsdk together with C++ template libraries
cpprestsdk is defining a U macro that breaks when combined with many popular C++ libraries: template parameters are usually named T, U, V. This includes Boot, and Intel TBB amongst others.

While the U macro is somewhat convenient for you, it is a major inconvenience for users of the library. Does it make sense to reconsider using it?

Also: Related to closed issue #647

garethsb-sony:
…in order that SDK clients can build with ``_TURN_OFF_PLATFORM_STRING`` defined.

There are currently many instances of ``U()`` in src, samples and tests\functional, that mean it's not currently possible to build the SDK itself or many of the samples or tests with ``_TURN_OFF_PLATFORM_STRING`` defined. This PR doesn't address that.

 Fixes #692.

### DIFF
--- a/Release/include/cpprest/details/basic_types.h
+++ b/Release/include/cpprest/details/basic_types.h
@@ -80,6 +80,9 @@ typedef std::stringstream stringstream_t;
 #endif // endif _UTF16_STRINGS
 
 #ifndef _TURN_OFF_PLATFORM_STRING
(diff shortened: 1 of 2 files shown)


ras0219-msft:
Thanks!

PR #711, opened 2018-03-05
Update the build process for Boost on Android

jwtowner:
The version of the external Boost-for-Android build tool has been
brought up to date and the Boost.Build jam files that were
available for use with Boost 1.66.0 have been backported to Boost
1.55.0 and 1.59.0 currently used by cpprestsdk. Furthermore, they
have been modified to support the following features:

- Added support for building with all recent Android NDK releases,
  including r10e, r11c, r12b, r13b, r14b, r15c and r16b.
- Support for all current Android target architectures is now
  available and working (armeabi, armeabi-v7a, arm64-v8a, mips,
  mips64, x86 and x86_64). All architectures are handled by the
  same patch file, there is no longer any need to download and
  patch Boost-for-Android separately for each unique target
  architecture, which should speed up the build process.
- Added --stl and --api options to build-android.sh for selecting
  the C++ runtime library and Android API level to build against.
- Added --without-supportlib to disable use of Android Support
  Library header files, to maintain backward compatability with
  previous behavior.
- Building Boost within Ubuntu on WSL now works correctly, as some
  of the environment variables used for paths weren't correctly
  encapsulated in quotes in order to handle paths containing
  space characters.

All of these features are available for Boost versions 1.55.0, 1.59.0,
1.65.1 and 1.66.0.

The default build configuration for Boost on Android remains
unchanged from the current settings. Boost is built with the
gnustl_static C++ runtime library, Android-9 API level, without
use of the Android Support Library and only for x86 and
armeabi-v7a target architectures. However, this can now easily
be changed in configure.sh by passing in different arguments
to build-android.sh.


jwtowner:
I've added extra support for Boost 1.59.0 as that seems to be the actual precompiled version of Boost used for cpprestsdk on Android with the Xbox Live SDK.

PR #715, opened 2018-03-09
fix template whitespace syntax

aduskett:
Some files don't have a space inbetween the '<' and '::' charachters, which
will cause build failures on older toolchains. Adding a space inbetween these
two characters fixes the issue.

See http://autobuild.buildroot.net/results/797a9b5fdf6ab0f16f2249324b48292dfab61d9f/build-end.log
for more information.


garethsb-sony:
This may also be resolved by ``-fpermissive`` on GCC 4.7 as far as I remember. See comments in #609.

ras0219-msft:
Thank you for the PR, however note that we will not be testing with these old compilers in the future and make no guarantees that it won't be broken again :)

PR #722, opened 2018-03-23
Fixed proxy issue when using WPAD on Windows 8.1+

tbarnekov:
If IE was set to use Web-Proxy Auto-Dicovery (WPAD) on Windows 8.1 or newer, a request would fail with "Settings proxy options: 12011: The option value cannot be set" when trying to set the proxy for the request.

The new WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY doesn't allow manually specifying the proxy by using WinHttpSetOption.

Please note that the error is only triggered if the executable has a manifest specifying support for Windows 8.1+. Otherwise the calling process will run in Windows Vista compatibility mode, and not trigger the error.

### DIFF
--- a/Release/src/http/client/http_client_winhttp.cpp
+++ b/Release/src/http/client/http_client_winhttp.cpp
@@ -431,37 +431,40 @@ protected:
 #ifndef CPPREST_TARGET_XP
             if (IsWindows8Point1OrGreater())
             {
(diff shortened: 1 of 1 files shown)


scissorrunner:
When this will be merged to master?  I've encountered this issue as well, and would prefer not to have to run in Windows 7 compatibility mode.

ras0219-msft:
Thanks for the PR!

PR #732, opened 2018-04-10
Fix a build problem on Clang.

adah1972:
AND_CAPTURE_MEMBER_FUNCTION_POINTERS workaround had a check for GCC, but did not exclude Clang.  Clang has a fake GCC version of 4.2, thus caused problems.


ras0219-msft:
Thanks!

PR #737, opened 2018-04-19
Update CMakeLists.txt to install the cmake bindings in the right location

LocutusOfBorg:
Update CMakeLists.txt to install the cmake bindings in the right location
/usr/lib/cpprestsdk is not the correct FHS standard cmake location.
It should be placed in /usr/lib/cmake/cpprestsdk instead.


ras0219-msft:
Thanks for the PR and sorry for the long delay!

I agree that we should use `GNUInstallDirs` to enable configuration of the cpprestsdk bin/lib output directories, however we already have a way to configure the cmake config file output location. Therefore, I suggest using that instead to achieve your desired target:
```
cmake .. -DCPPREST_EXPORT_DIR=lib/cmake/cpprestsdk
```

PR #746, opened 2018-04-27
Fix compilation failure with ``CPPREST_FORCE_PPLX=1`` (resolves #734)

### LINKED ISSUE #734: Compiling with CPPREST_FORCE_PPLX for WIN32 in vcpkg
Hi,

I am trying to compile a win32 package of cpprestsdk with vcpkg and enforce the use of pplx.

I added the -DCPPREST_PPLX_IMPL:STRING=winpplx variable to the portfile.cmake, so that the config block looks like:
```
vcpkg_configure_cmake(
    SOURCE_PATH ${SOURCE_PATH}/Release
    PREFER_NINJA
    OPTIONS
        ${OPTIONS}
        -DBUILD_TESTS=OFF
        -DBUILD_SAMPLES=OFF
        -DCPPREST_EXCLUDE_WEBSOCKETS=OFF
        -DCPPREST_EXPORT_DIR=share/cpprestsdk
        -DWERROR=OFF
        -DCPPREST_PPLX_IMPL:STRING=winpplx
    OPTIONS_DEBUG
        -DCPPREST_INSTALL_HEADERS=OFF
)
```
The build will fail with several errors.

Compiling in vcpkg without the variable works. Also compiling from the visual strudio project files from this repository works, when the CPPREST_FORCE_PPLX preprocessor definition is set.

What am I doing wrong?

garethsb-sony:
See #734 for some analysis.

In summary: make sure internal ``_do_while`` fucntion is in the same namespace as other pplx tasks implementation details.

Original problem reproduced and this fix tested with:

* Visual Studio 2013 (the OP reported on 2015)
* Boost 1.66.0 (also applies to Boost 1.67.0, however Boost 1.65.1 does not repro)
* ``CPPREST_FORCE_PPLX=1``

(I don't know why Release/include/cpprest/astreambuf.h is showing up as a large diff, my local diff shows only +3~2. Line endings issue maybe?)


ras0219-msft:
Thanks LGTM! I also fixed the line endings.

Issue #742, opened 2018-04-24
hostport_listener stop listening for connections after first call to on_accept with an error


sridmad:
The code in hostport_listener::on_accept does not subscribe to listening for new connections if it was called with an error. As a result a http based rest server stops accepting connections if an error occurs.

See https://www.boost.org/doc/libs/1_63_0/doc/html/boost_asio/tutorial.html#boost_asio.tutorial.tutdaytime3 for pattern.

garethsb-sony:
This seems true. I'm surprised it's not been reported before. Do you have an example or test that reproduces such an error happening?

sridmad:
We have hit this issue on and off in our product and found the issue through code review and confirmed through logging. Unfortunately no consistent repro. 

I have made the change in a fork. All Debug tests pass. Some tests fail in Release build but not consistently (i.e. different tests are failing on different runs). Not sure if it is due to the change or the release tests have the issues. 

Will take a look at the test failures to see if they have a dependency on the existing pattern.

garethsb-sony:
That would be great.

Have you logged which ``error_code`` values are seen?

sridmad:
Error code was not logged previously. We have since added code to log the error code to cpprestsdk and rebuilding our product. We should have the error code value(s) in a day or two.

garethsb-sony:
This looks strongly related to #575 which does have a repro and a suggested fix. I haven't confirmed it myself.

sridmad:
@garethsb-sony, #546  seems to handle one of the cases. I suspect there are other test cases that has to be handled.

https://github.com/chriskohlhoff/asio/blob/ad1751063aa717bd128dddc23456667ab816b2d9/asio/src/examples/cpp03/http/server/server.cpp seems to handle socket close case.

PR #750, opened 2018-05-04
Improve error handling in on_accept

sridmad:
1. Handle operation_abort error code. Indicates that the listen socket is closed

2. Handle exceptions thrown by start. Setting ssh context can throw exceptions, catch them, close the connection and continue handling connections.

3. Continue accepting new connections for any error_code other than operation_abort.

### DIFF
--- a/Release/src/http/listener/http_server_asio.cpp
+++ b/Release/src/http/listener/http_server_asio.cpp
@@ -540,27 +540,43 @@ will_deref_and_erase_t asio_server_connection::start_request_response()
 
 void hostport_listener::on_accept(ip::tcp::socket* socket, const boost::system::error_code& ec)
 {
-    std::unique_ptr<ip::tcp::socket> usocket(std::move(socket));
+    // Listener closed
+    if (ec == boost::asio::error::operation_aborted)
+    {
+        return;
+    }
 
+    // Handle successfull accept
     if (!ec)
     {
+        std::unique_ptr<ip::tcp::socket> usocket(std::move(socket));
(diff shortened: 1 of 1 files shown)


sridmad:
@kavyako , please review

sridmad:
Addresses #742

garethsb-sony:
Is it possible to receive ``operation_aborted`` for a temporary error condition?

sridmad:
@garethsb-sony, I found a different mechanism to detect connection closed in a asio sample at https://github.com/chriskohlhoff/asio/blob/master/asio/src/examples/cpp03/http/server/server.cpp. However this pattern does not handle all cases correctly. There are cases where on_accept is with operation_aborted before m_acceptor.is_open() is able to detect that the connection is closed. Most probably due to a race in the shutdown case.

My change will not regress existing behavior but at least allows the server to continue listening on other errors while having the same behavior for operation_aborted. I was not and expert on Linux and was not able to find much information about other cases that can result in operation_aborted.

sridmad:
@garethsb-sony, without the lock, m_acceptor can be reset between the null check and using the variable.

garethsb-sony:
Thanks. OK - yes, m_acceptor is reset under the mutex in hostport_listener::stop. Now I'm wondering whether hostport_listener::start should also take the lock in order to set m_acceptor.

sridmad:
@garethsb-sony, I cannot see anyone using an instance of hostport_listener before Start completes. For all practical purposes Start is the constructor for the type having multiple threads calling Start on the same instance is a bug. Existing calls to Start on this type all ensure that Start is called from only one thread and other members are not called unless Start has completed.

PR #752, opened 2018-05-05
Require at least CMake 3.1 due to use of target_sources

### LINKED ISSUE #633: CMake minimum required version should actually be 3.1
`Release/CMakeLists.txt` says `cmake_minimum_required(VERSION 3.0)`.
But `Release/src/CMakeLists.txt` makes use of the `target_sources` command which is only available
in CMake 3.1 and above.

vadz:
target_sources() is only available since this version and this avoids

CMake Error at src/CMakeLists.txt:92 (target_sources):
  Unknown CMake command "target_sources".

when using CMake 3.0, for example.

Closes #633.


ras0219-msft:
LGTM, thanks!

PR #762, opened 2018-05-16
Fix handling of timed out connections kept alive in connection pool under Unix

### LINKED ISSUE #592: http_client does not gracefully handle remotely-closed keepalive connections (Linux/ASIO)
From my discussion with Robert, the issue manifests itself like so:

1. Have a keepalive socket as an `http_client` to some server.
2. Server sends a FIN ACK as it decides the connection has been idle for too long which puts the socket into CLOSE_WAIT
3. something comes along and makes a new request with the `http_client` instance, headers are sent on the socket which is in CLOSE_WAIT so they are still received.
4. Since the server can't actually respond, the connection gets an RST from the server end and cpprestsdk finds out about it when it moves onto trying to write the body, consequently throwing an exception to the user.

Firstly, we should try other connections if we end up in this state, however, retrying all connections will result in one unlucky caller spending time acting as a garbage collector for all the dead sockets, which could be numerous.

One option for dealing with this is to retry the pool of existing connections in a limited fashion such that anyone who encounters 2 failures will give up and (try to) establish a new connection -  so they'll still have cleaned up more than they created.

vadz:
This fixes #592 by ensuring that we actually reopen the connection if the existing one, obtained from the connection pool, timed out in the meanwhile.

Note that the existing code already tried to do it, but failed due to 2 bugs in it: first, it didn't detect the timeouts properly for HTTPS connections and, second, it didn't reinitialize the request input stream, meaning that even if the connection was reopened, the correct request wasn't sent over it.

Other improvements remain possible (and desirable), but this PR at least makes things work, albeit not in the most efficient way, whereas they just don't currently, so I think it should already be applied and the rest done later.


ras0219-msft:
Agreed. Thanks for the PR and sorry for the long delay!

Issue #773, opened 2018-06-04
Compiling for iOS using Xcode 9 tool chain has lambda capture 'this' is not used error


mobileben:
In http_client_asio.cpp I am getting this error:

>  error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
>         auto start_http_request_flow = [proxy_type, proxy_host, proxy_port AND_CAPTURE_MEMBER_FUNCTION_POINTERS](std::shared_ptr<asio_context> ctx)

Clang compiler also sets __GNUC__, so inevitably this get set

`#define AND_CAPTURE_MEMBER_FUNCTION_POINTERS , this`

Currently I am doing this

`#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8) || (__clang__ && __APPLE__)`

As a work around. If this seems reasonable or if I should just drop the `&& __APPLE__`, let me know.

I am currently updating the build process for iOS, so I plan on putting this change in that pull request.

garethsb-sony:
See PR #732 (probable simple fix you could check) and issue #768 (workaround).

@ras0219-msft please would you review that PR?

mobileben:
@garethsb-sony @ras0219-msft 

Hmm, didn't see that one. More or less same approach to fix, just different locations and I added __APPLE__ which was probably not needed, but didn't want to step on anyone's toes.

I probably won't submit my PR for iOS today, since I'm declaring bankruptcy on my older PR for it and am re-doing it (small changes so no biggie). Hopefully #732 will get approved....

aksswami:
@ras0219-msft  Can we get this one merged?

Issue #770, opened 2018-05-28
Make for iOS platform no longer working. Giving arch error as armv7 and armv7s removed from > iOS 11.0


aksswami:
I was trying to build the cpprestsdk for iOS platform. But getting below error 
```
clang: error: invalid iOS deployment version '-mios-simulator-version-min=11.2', iOS 10 is the maximum deployment target for 32-bit targets [-Winvalid-ios-deployment-target]
```

`cmake.toolchain` file needs to be updated for latest changes in iOS 11.0 . As I am very new to CMake and C++, not sure where and what changes to do. 

Any help would be really helpful. Thanks.

mobileben:
@aksswami I created a PR that should help you past this.

https://github.com/Microsoft/cpprestsdk/pull/776

aksswami:
Thanks @mobileben.

I was able to build this using some other toolchain for ios cmake and editing cmake file for iOS. But as I am new to CMake not very sure if the changes were correct. This will really help as build process and other parts are out of date on this.

Thanks again!

mobileben:
@aksswami most toolchains for iOS are about the same. Although I don't believe they all work correctly  since toolchains cannot access variables set outside of the tool chain (for the most part). 

This was why I bridged over the deployment version through an environment variable. Not my first choice, but I could not get global properties to work right.

I am by no means a cmake pro ... rather I can do the basics. So it could also be it I was using the tooling incorrectly.

Much of the iOS stuff is out of date or partially described.

It is best to build boost from here

https://github.com/faithfracture/Apple-Boost-BuildScript

If you want static libs (which is what I use), I have a PR there to build a universal fat binary (or you can go to my fork).

I have a hacky script to build openssl here. It really isn't meant to be iron clad. It does what I need it to do so I haven't refined it to be more forgiving.

once those pieces are in place, then it's very buildable (using my PR).

https://gist.github.com/mobileben/dbe72732d9446a1a055f9d7fb453464c

aksswami:
Thanks for the info. I am using this https://github.com/krzyzanowskim/OpenSSL for OpenSSL on both OSX and iOS and things are working fine, as I just tested your PR.

mobileben:
My script also builds MacOS (or rather should). Definitely use what works best for your situation.

PR #776, opened 2018-06-06
#436 Building iOS doesn’t seem to work

mobileben:
My previous PR https://github.com/Microsoft/cpprestsdk/pull/439 had a conflict.

This is updated to work with issues related to Xcode 9. This should also address

https://github.com/Microsoft/cpprestsdk/issues/436
https://github.com/Microsoft/cpprestsdk/issues/773 (note another PR addresses this too)
https://github.com/Microsoft/cpprestsdk/issues/770

Sample cmake to set deployment is 

cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_XCODE_ATTRIBUTE_IPHONEOS_DEPLOYMENT_TARGET=9.3 ..

I also remove -fvisibility-inlines-hidden as Xcode defaults it to OFF.

iOS.cmake is now part of this. The recommended toolchain has not been modified in a while. I forked it here.

https://github.com/mobileben/ios-cmake

And this also allows usage of static boost or framework boost.


aksswami:
Tried this today and working great. Thanks @mobileben

ras0219-msft:
Thanks for the PR @mobileben and sorry for taking so long to address it!

PR #779, opened 2018-06-06
[gcc-8][clang++-6] disable more -Werror warnings

moriarty:
gcc-8: -Wno-format-truncation

clang-6: -Wdelete-non-virtual-dtor
clang-6: -Wunused-lambda-capture

removed duplicated: -Wno-reorder

This fixes #778

Issue #778, opened 2018-06-06
[gcc-8][clang-6][-Werror] build fails due to warnings.


moriarty:
On latest master, building with GCC-8 on Ubuntu 18.04. 

Build fails due to **-Werror=format-truncation** 

**cmake info**
```
cmake ..
-- Setting gcc options
-- Found websocketpp version 0.7.0 on system
-- Boost version: 1.65.1
-- Found the following Boost libraries:
--   random
--   system
--   thread
--   filesystem
--   chrono
--   atomic
--   date_time
--   regex
-- Added test library httpclient_test
-- Added test library httplistener_test
-- Added test library json_test
-- Added test library pplx_test
-- Added test library streams_test
-- Added test library uri_test
-- Added test library utils_test
-- Added test library websocketsclient_test
-- Configuring done
-- Generating done
-- Build files have been written to: /home/amoriarty/workspace/microsoft/cpprestsdk/Release/build
```

**make output**
```
/home/amoriarty/workspace/microsoft/cpprestsdk/Release/src/utilities/asyncrt_utils.cpp: In member function ‘utility::string_t utility::datetime::to_string(utility::datetime::date_format) const’:
/home/amoriarty/workspace/microsoft/cpprestsdk/Release/src/utilities/asyncrt_utils.cpp:702:42: error: ‘%s’ directive output may be truncated writing up to 8 bytes into a region of size between 1 and 65 [-Werror=format-truncation=]
         snprintf(output, sizeof(output), "%s%sZ", datetime_str, buf);
                                          ^~~~~~~                ~~~
/home/amoriarty/workspace/microsoft/cpprestsdk/Release/src/utilities/asyncrt_utils.cpp:702:17: note: ‘snprintf’ output between 2 and 74 bytes into a destination of size 65
         snprintf(output, sizeof(output), "%s%sZ", datetime_str, buf);
         ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
src/CMakeFiles/cpprest.dir/build.make:446: recipe for target 'src/CMakeFiles/cpprest.dir/utilities/asyncrt_utils.cpp.o' failed
```

**quick fix**

Just ignore those warnings too with **-Wno-format-truncation** There are a lot of warnings disabled (some duplicates) already in the **CMakelists.txt** ... some of these might be masking actual issues so this is just a quick fix.
[...]

PR #787, opened 2018-06-17
Fix gcc8 error/warning -Werror=format-truncation=

andreasstieger:
utilities::datetime::to_string(): datetime_str and buf were oversized
for fitting into output without possible trunctation


andreasstieger:
This actually fixes part of #778, and at least for this warning is better than #779.

ras0219-msft:
Agreed, this is a good fix. Thanks for the PR!

PR #809, opened 2018-07-16
Fix for compilation with MSVC permissive-

robert-andrzejuk:
Related Visual Studio issue:
https://developercommunity.visualstudio.com/content/problem/238445/template-class-cant-access-protected-member-when-c.html

Problem is with code compiling with /permissive-, so the template functions are affected by this issue.


ras0219-msft:
Thanks for the PR!

PR #814, opened 2018-07-23
Update asyncrt_utils.cpp to fix build failure with gcc-8

LocutusOfBorg:
output size should at least equal to sizeof(datetime_str) + sizeof(buf)
/cpprest/Release/src/utilities/asyncrt_utils.cpp: In member function 'utility::string_t utility::datetime::to_string(utility::datetime::date_format) const':
/cpprest/Release/src/utilities/asyncrt_utils.cpp:702:42: error: '%s' directive output may be truncated writing up to 8 bytes into a region of size between 1 and 65 [-Werror=format-truncation=]
         snprintf(output, sizeof(output), "%s%sZ", datetime_str, buf);
                                          ^~~~~~~                ~~~
In file included from /usr/include/stdio.h:862,
                 from /usr/include/c++/8/cstdio:42,
                 from /usr/include/c++/8/ext/string_conversions.h:43,
                 from /usr/include/c++/8/bits/basic_string.h:6391,
                 from /usr/include/c++/8/string:52,
                 from /cpprest/Release/src/pch/stdafx.h:50,
                 from /cpprest/Release/src/utilities/asyncrt_utils.cpp:14:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:35: note: '__builtin___snprintf_chk' output between 2 and 74 bytes into a destination of size 65
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors

### DIFF
--- a/Release/src/utilities/asyncrt_utils.cpp
+++ b/Release/src/utilities/asyncrt_utils.cpp
@@ -685,7 +685,7 @@ utility::string_t datetime::to_string(date_format format) const
     gmtime_r(&time, &datetime);
 
     const int max_dt_length = 64;
-    char output[max_dt_length+1] = {0};
+    char output[max_dt_length+10] = {0};
 
     if (format != RFC_1123 && frac_sec > 0)
     {


LocutusOfBorg:
probably a better fix is #787

Issue #813, opened 2018-07-23
Build error with boost 1.63 (error converting from boost::system::error_code to td::error_code)


ashishsb95:
**Bug**
Casablanca/Release/libs/websocketpp/websocketpp/transport/asio/security/tls.hpp:315:39: error: could not convert ‘boost::asio::ssl::error::make_error_code((boost::asio::ssl::error::stream_errors)1)’ from ‘boost::system::error_code’ to ‘std::error_code’
                 return make_error_code(boost::asio::ssl::error::stream_truncated);

**system specifications**
os = fedora 26
c++ compiler = g++ (GCC) 7.3.1 20180130 (Red Hat 7.3.1-2)
ssl = openssl   1:1.1.0
zlib 1.2.11-2

**Fix**
`if (ERR_GET_REASON(ec.value()) == boost::asio::ssl::error::stream_truncated) {
                return make_error_code(static_cast<std::errc>(boost::asio::ssl::error::stream_truncated));`

LocutusOfBorg:
@pvazteixeira please install libwebsocketpp-dev package, to avoid using the bundled version!

PR #815, opened 2018-07-23
Fix boost1.63 build failure Fixes: #813

LocutusOfBorg:


ras0219-msft:
Thanks for the PR; however I thought that on Debian the system was providing the copy of websocketspp?

LocutusOfBorg:
Yes, you are right, in both Debian and Yocto, this patch is useless.
Sometimes I do testing without installing external dependencies, so I get errors when using bundled dependencies... (and according to #813) I'm not alone :dagger: 
is it possible to sync websocketpp to 0.8.1?

Issue #849, opened 2018-08-22
no response when I used wrk to test the server based on cpprest


winstard:
I changed the source code of CasaLens(the sample project) like this:
```
 void CasaLens::handle_get(http_request message)
{    
	static std::mutex mutex;
	message.reply(web::http::status_codes::OK);
	std::lock_guard<std::mutex>` `lock(mutex);
	static int b = 0;
	std::wcout << "b " << b++ << std::endl;
	return;
}
```
Then I  used wrk to test the performance 
`wrk -c 1500 -t 10 -d 1m --timeout=30s http://10.22.99.13:12344`

the result was:
```
Running 1m test @ http://10.22.99.13:12344
  10 threads and 1500 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.24s   284.07ms   1.42s    86.96%
    Req/Sec    13.17     14.51    42.00     83.33%
  23 requests in 1.00m, 2.38KB read
  Socket errors: connect 489, read 169, write 0, timeout 0
Requests/sec:      0.38
Transfer/sec:      40.57B
```

It just printed 
```
b 1
b 2
...
b 1033
```
on the consle of the CasaLens.
And there was no response when I hit Enter in the consle.

I tried to debug the cpprest. Found that it can't callback after called `HttpReceiveHttpRequest` at `void windows_request_context::async_process_request(HTTP_REQUEST_ID request_id, http_request msg, const unsigned long headers_size)`,when wrk has been executed for a while.
And no response after `listener.close().wait();`

The same question when I try to use cpprest to create a restful service on Linux.
Is my usage or something else wrong?

PR #844, opened 2018-08-18
Fix another clang build failure (Fixes: #747)

LocutusOfBorg:
===>  Building for cpprestsdk-2.9.1_5
[1/161] /usr/local/libexec/ccache/c++  -Dcpprest_EXPORTS -Iinclude -isystem /usr/local/include -isystem libs/websocketpp -Isrc/../include -Isrc/pch -O2 -pipe -fstack-protector -fno-strict-aliasing -stdlib=libc++ -Wno-return-type-c-linkage -Wno-unneeded-internal-declaration -std=c++11 -fno-strict-aliasing -O2 -pipe -fstack-protector -fno-strict-aliasing -fPIC   -Wall -Wextra -Wcast-qual -Wconversion -Wformat=2 -Winit-self -Winvalid-pch -Wmissing-format-attribute -Wmissing-include-dirs -Wpacked -Wredundant-decls -Wno-overloaded-virtual -Wno-sign-conversion -Wno-deprecated -Wno-unknown-pragmas -Wno-reorder -Wno-char-subscripts -Wno-switch -Wno-unused-parameter -Wno-unused-variable -Wno-deprecated -Wno-unused-value -Wno-unknown-warning-option -Wno-return-type-c-linkage -Wno-unused-function -Wno-sign-compare -Wno-shorten-64-to-32 -Wno-reorder -Wno-unused-local-typedefs -Werror -pedantic -MD -MT src/CMakeFiles/cpprest.dir/pplx/pplx.cpp.o -MF src/CMakeFiles/cpprest.dir/pplx/pplx.cpp.o.d -o src/CMakeFiles/cpprest.dir/pplx/pplx.cpp.o -c src/pplx/pplx.cpp
FAILED: src/CMakeFiles/cpprest.dir/pplx/pplx.cpp.o
[...]

PR #923, opened 2018-10-17
Re-enable no-unused-lamda-capture and no-format-trucation

moriarty:
These were disabled as a quick fix for #778 in #779.

* no-format-truncation was properly fixed in #787
* no-unused-lamda-capture was worked around in #732
* gaurds for gcc 5.4 were added in #849 but with #787 aren't needed

I did the original quick fix of disabling the warnings, and I meant to come back and fix them when I had time.


moriarty:
I didn't re-enable the **-Wdelete-non-virtual-dtor** because at first I couldn't find when it had been fixed.
But it was fixed with #844.

It looks like this repo uses squash-and-merge so I will push a new commit to this branch.

PR #1143, opened 2019-05-27
add ping and pong to message handler

Kumamon38:
using `set_ping_hanlder`, `set_pong_handler`


Release/src/websockets/client/ws_client_wspp.cpp
```diff
@@ -333,7 +333,8 @@ class wspp_callback_client : public websocket_client_callback_impl,
                     websocket_incoming_message incoming_msg;
 
                     incoming_msg.m_msg_type = websocket_message_type::ping;
-                    incoming_msg.m_body = concurrency::streams::container_buffer<std::string>(msg);
+                    // 'move' the payload into a container buffer to avoid any copies.
+                    incoming_msg.m_body = concurrency::streams::container_buffer<std::string>(std::move(msg));
```
Dimon4eg: `std::move` is redundant here. Compiler will ignore it because msg is const ref. To use `std::move` the msg should be &&:
`std::string&& msg`

Release/include/cpprest/ws_msg.h
```diff
@@ -64,6 +64,12 @@ class websocket_outgoing_message
     /// This is useful when the client side wants to check whether the server is alive.
     /// </summary>
     void set_pong_message() { this->set_message_pong(); }
+
```
garethsb-sony: Just a suggestion: put `set_ping_message` above `set_pong_message`.

Release/include/cpprest/ws_msg.h
```diff
@@ -64,6 +64,12 @@ class websocket_outgoing_message
...
+    /// <summary>
+    /// Sets a the outgoing message to be a ping message.
```
garethsb-sony: Copy-and-pasted typo, change 'a the' to 'the'.

Release/include/cpprest/ws_msg.h
```diff
@@ -64,6 +64,12 @@ class websocket_outgoing_message
     /// This is useful when the client side wants to check whether the server is alive.
```
garethsb-sony: Nice to fix old typo, change 'a the' to 'the'.
Also, I would delete the sentence 'This is useful...' because I don't think it's true. Unsolicited pongs get no response. It *is* true for pings, so makes sense where you have copied it to in the new `set_ping_message`!

Release/include/cpprest/ws_msg.h
```diff
@@ -147,6 +153,13 @@ class websocket_outgoing_message
         m_length = static_cast<size_t>(buffer.size());
         m_body = buffer;
     }
+    void set_message_ping()
```
garethsb-sony: Just a suggestion: put `set_message_ping` above `set_message_ping`.

Release/include/cpprest/ws_msg.h
```diff
@@ -64,6 +64,12 @@ class websocket_outgoing_message
     /// This is useful when the client side wants to check whether the server is alive.
     /// </summary>
     void set_pong_message() { this->set_message_pong(); }
+
+    /// <summary>
+    /// Sets a the outgoing message to be a ping message.
+    /// This is useful when the client side wants to check whether the server is alive.
+    /// </summary>
+    void set_ping_message() { this->set_message_ping(); }
```
garethsb-sony: Suggestion: add optional application-data argument, i.e. `const std::string& data = ""`, since ping/pong messages _may_ have bodies up to ~125 bytes. If so, should also be added to `set_pong_message`.

Release/src/websockets/client/ws_client_wspp.cpp
```diff
@@ -694,6 +724,7 @@ class wspp_callback_client : public websocket_client_callback_impl,
             case websocket_message_type::binary_message:
                 client.send(this_client->m_con, sp_allocated.get(), length, websocketpp::frame::opcode::binary, ec);
                 break;
+            case websocket_message_type::ping: client.ping(this_client->m_con, "", ec); break;
```
garethsb-sony: Use message data, if added as suggested above.

Release/include/cpprest/ws_msg.h
Kumamon38: done

Release/include/cpprest/ws_msg.h
Kumamon38: changed too

Release/include/cpprest/ws_msg.h
Kumamon38: inverted too

Release/include/cpprest/ws_msg.h
Kumamon38: Added string data param

Release/src/websockets/client/ws_client_wspp.cpp
Kumamon38: Added
As websocketpp doesn't have a `ping(void*, length)` but just a `ping(std::string param)`, we need to cast

Kumamon38:
Btw I added a `set_auto_pong_enabled()`in the config in order to automatically answer to ping requests

Release/src/websockets/client/ws_client_wspp.cpp
```diff
@@ -724,8 +724,8 @@ class wspp_callback_client : public websocket_client_callback_impl,
             case websocket_message_type::binary_message:
                 client.send(this_client->m_con, sp_allocated.get(), length, websocketpp::frame::opcode::binary, ec);
                 break;
-            case websocket_message_type::ping: client.ping(this_client->m_con, "", ec); break;
-            case websocket_message_type::pong: client.pong(this_client->m_con, "", ec); break;
+            case websocket_message_type::ping: client.ping(this_client->m_con, *(std::string*)sp_allocated.get(), ec); break;
```
garethsb-sony: Is that cast OK? Isn't `sp_allocated` a pointer to raw bytes, not a `std::string` object?
Since `ping` and `pong` only accept `std::string` we probably need to construct one here something like: `std::string(sp_allocated.get(), length)`?

Release/src/websockets/client/ws_client_wspp.cpp
```diff
@@ -337,7 +337,7 @@ class wspp_callback_client : public websocket_client_callback_impl,
 
                     m_external_message_handler(incoming_msg);
                 }
-                return true;
+                return m_config.is_auto_pong_enabled();
```
garethsb-sony: [RFC 6455 Section  5.5.2](https://tools.ietf.org/html/rfc6455#section-5.5.2) basically says that you MUST send a pong in answer to a ping. Section 5.5.3 says that if you receive multiple pings before you get a chance to send a pong, you MAY only acknowledge the last one. However supporting that behaviour would be more complicated than just having a Boolean in config.

Therefore I suggest _not_ adding `is_auto_pong_enabled` in this PR.

Release/include/cpprest/ws_msg.h
```diff
@@ -60,16 +60,23 @@ class websocket_outgoing_message
 public:
 #if !defined(__cplusplus_winrt)
     /// <summary>
-    /// Sets a the outgoing message to be an unsolicited pong message.
-    /// This is useful when the client side wants to check whether the server is alive.
```
garethsb-sony: It looks like you have deleted this sentence from both `set_ping_message` and `set_pong_message`. I meant that you should keep this sentence only for `set_ping_message`, please.

Release/src/websockets/client/ws_client_wspp.cpp
Kumamon38: Yeah but we need to return true or false here, and before we didn’t acknowledge the ping by a pong at all as we had no callback yet; so we need to add it

Release/src/websockets/client/ws_client_wspp.cpp
Kumamon38: I’ll change that

Release/src/websockets/client/ws_client_wspp.cpp
garethsb-sony: I have found that websocketpp sends pong in response to ping by default, and will do so if you just return `true`, so I would stick with doing that, as you had originally.

garethsb-sony:
I've added some unit tests for this to https://github.com/garethsb-sony/cpprestsdk/commit/468ce717df7fe1c96eb2286f40ed7e3f2e970e8b. GitHub won't let me submit a PR on your branch.

Kumamon38:
Not sure why I doesn't accept PR but anyway I added your commits :)

Release/include/cpprest/ws_msg.h
```diff
@@ -60,10 +60,23 @@ class websocket_outgoing_message
 public:
 #if !defined(__cplusplus_winrt)
     /// <summary>
-    /// Sets a the outgoing message to be an unsolicited pong message.
+    /// Sets the outgoing message to be a ping message.
     /// This is useful when the client side wants to check whether the server is alive.
     /// </summary>
-    void set_pong_message() { this->set_message_pong(); }
+    /// <param name="data">UTF-8 String containing the optional ping message.</param>
+    void set_ping_message(const std::string& data = "")
```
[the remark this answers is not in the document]
garethsb-sony: A ping message may or may not have a payload. But you still need a way to make the outgoing message into a ping message?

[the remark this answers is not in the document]
BillyONeal: I see. In that case it is more efficient to use `std::string()` for this than `""`.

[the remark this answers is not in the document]
Kumamon38: I’ll change tomorrow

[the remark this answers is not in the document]
garethsb-sony: Just as efficient if not more so, depending on compiler, would be to use `{}` hich avoids repeating the type.

https://godbolt.org/z/ausWRL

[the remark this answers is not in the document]
BillyONeal: > Just as efficient if not more so, depending on compiler, would be to use `{}` which avoids repeating the type.
> 
> https://godbolt.org/z/ausWRL

(1) That shows the `string()` version being more efficient, since it's pushing the "" constant on the stack in the caller. But that is with the optimizer off so the results don't super matter for that, we're talking 'small cheese'.

(2) If you include the caller (which is all that matters when you're talking about default arguments) and turn on the optimizer either string() or {} (both of which call std::string's default constructor) crush "": https://godbolt.org/z/bjGgME

```
$T1 = 32
__$ArrayPad$ = 64
unsigned __int64 meow2(void) PROC                           ; meow2, COMDAT
$LN104:
        sub     rsp, 88                             ; 00000058H
        mov     rax, QWORD PTR __security_cookie
        xor     rax, rsp
        mov     QWORD PTR __$ArrayPad$[rsp], rax
        xor     r8d, r8d
        mov     QWORD PTR $T1[rsp+16], 0
        lea     rdx, OFFSET FLAT:`string'
        mov     QWORD PTR $T1[rsp+24], 15
        lea     rcx, QWORD PTR $T1[rsp]
        mov     BYTE PTR $T1[rsp], 0
        call    std::basic_string<char,std::char_traits<char>,std::allocator<char> > & std::basic_string<char,std::char_traits<char>,std::allocator<char> >::assign(char const * const,unsigned __int64) ; std::basic_string<char,std::char_traits<char>,std::allocator<char> >::assign
        mov     rdx, QWORD PTR $T1[rsp+24]
        cmp     rdx, 16
        jb      SHORT $LN74@meow2
        mov     rcx, QWORD PTR $T1[rsp]
        inc     rdx
        mov     rax, rcx
        cmp     rdx, 4096               ; 00001000H
        jb      SHORT $LN84@meow2
        mov     rcx, QWORD PTR [rcx-8]
        add     rdx, 39                             ; 00000027H
        sub     rax, rcx
        add     rax, -8
        cmp     rax, 31
        ja      SHORT $LN102@meow2
$LN84@meow2:
        call    void operator delete(void *,unsigned __int64)               ; operator delete
$LN74@meow2:
        mov     rax, QWORD PTR $T1[rsp+16]
        mov     rcx, QWORD PTR __$ArrayPad$[rsp]
        xor     rcx, rsp
        call    __security_check_cookie
        add     rsp, 88                             ; 00000058H
        ret     0
$LN102@meow2:
        call    _invalid_parameter_noinfo_noreturn
        int     3
$LN100@meow2:
unsigned __int64 meow2(void) ENDP                           ; meow2
```

for "" vs.

```
unsigned __int64 purr2(void) PROC                           ; purr2, COMDAT
        xor     eax, eax
        ret     0
unsigned __int64 purr2(void) ENDP                           ; purr2
```

(I am the std::basic_string maintainer for MSVC -- I can tell you way more about that than about any of this HTTP stuff :) )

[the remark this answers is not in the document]
BillyONeal: Sorry I first read your response as saying "`""` is just as efficient" :P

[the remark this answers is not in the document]
Kumamon38: So what do we choose finally? ^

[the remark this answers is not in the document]
BillyONeal: Either `string()` or `{}` both call the `basic_string::basic_string()` constructor and are equivalent. `""` calls the `basic_string::basic_string(const CharT*)` constructor which is less efficient (there's a strlen and a memory allocation in there; even though for "" it won't ever actually allocate it's difficult for optimizers to prove this unless the universe happens to get inlined).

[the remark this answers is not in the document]
Kumamon38: Ok then we can let {}

BillyONeal:
Thanks for your contribution!

Issue #72, opened 2016-02-13
[feature request] std::string support for json library


petke:
In my experience most 3rd party libraries use std::string and not std::wstring in their interface. Therefore applications become tied to std::string. Because the ppl/json library uses std::wstring (on windows) this makes integrating the ppl/json library into an applications cumbersome and inefficient. One has to do lots of unnecessary wstring-to-string and string-to-wstring conversions. Could you add support for std::string, at least for the json library?

Minimal example:

		#include <cpprest/json.h>

		std::string SerializeToJson(std::string name, int id) //utf8 coming in and goin out
		{
			utility::stringstream_t stream;
			web::json::value js_obj;
			js_obj[L"id"] = web::json::value::number(id);
			js_obj[L"name"] = web::json::value::string(utility::conversions::utf8_to_utf16(name)); //unneccessary conversion 1

			js_obj.serialize(stream);
			std::wstring wret = stream.str();
			std::string ret = utility::conversions::utf16_to_utf8(wret); //unneccessary conversion 2

			return ret;
		}

		int main() {

			auto r = SerializeToJson("test", 123);
			std::cout << r << std::endl;
		}

jameskennedy2265:
I would love to see this as a feature as well. Also huge thanks to petke for showing how to get a string variable into a json container.

While you may not be able to update this as a feature, can anyone update the documentation with this in json, it would be a huge help to a lot of people including myself.

ras0219-msft:
I definitely agree that the wstring/string/string_t issues cause a ton of headache when using the library.

Let me lay out the current issues I see that need to be solve here.

- **On Windows, the underlying system APIs are utf16.** We want to give applications that use utf16 everywhere the option to avoid conversions (the current implementation handles this).
- **Virtually all web payloads are utf8.** Applications should be able to consume these payloads (including JSON) without incurring a utf8 -> utf16 hit (current implementation does _not_ allow this on windows unless you grab the data as a `vector<unsigned char>` and parse it yourself using RapidJSON/other).
- **Unfortunately, the default standpoint on windows is that `const char*`/`std::string` is encoded according to the local codepage**. For ASCII English this distinction of course doesn't matter, but we all need more ☃ in our lives. If possible, protection against encoding errors would be nice.
- **Don't break API-compat in minor versions.** If the solution requires breaking source-API compatibility, we wouldn't want to incorporate it until 3.0. I'd much prefer an incremental solution that we can release sooner!
[...]

PR #1199, opened 2019-07-19
add option to directly support narrow strings on win32

jason-ha:
add option to directly support narrow strings on win32
 independent of UNICODE vs MBCS setting
Set CPPREST_FORCE_NARROW_STRINGS=ON to enable.

Simplistic cleanup throughout project is not possible as some Windows APIs require utf16 such as HTTPSPolicyCallbackData in http_client_winhttp.cpp.
HTTP related code is not completely updated so a full package may not be produced, though settings exist to have a clean build. Use:
 -DCPPREST_FORCE_NARROW_STRING=ON -DCPPREST_EXCLUDE_WEBSOCKETS=ON -DCPPREST_HTTP_LISTENER_IMPL=none -DBUILD_SAMPLES=OFF


BillyONeal:
Look like legitimate build/test failures that need to be fixed.

jason-ha:
Indeed. Looks like simple header that needs added after my split. Been trying to wrap up the component isolation changes which can help find more like that. Will fix break this morning.

Jason

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

BillyONeal:
Unfortunately I don't think it is appropriate to merge this work with this still disabled, as merging this to master is making a statement that it's an officially supported part of the product. However, I acknowledge trying to make changes of this scale is difficult to do as a complete fully formed change if it were to be squashed as a single commit, and doing it as separate reviews certainly makes it easier to review. Are you OK with retargeting this change to a semi-supported branch for this, which we will merge with a non-squash commit later?

I reviewed this and agree with all the changes herein (other than the above concern).

jason-ha:
I wouldn't say it is disabled as it is an option for Win32. Most Win32 users will not want to enable it.
But option enable does not allow fully codebase use - I don't know from issue thread #72 whether anyone is asking for narrow strings for the other components.
Given that I see a few options:
 1) as is, but add yaml step to ensure it builds - maybe that doesn't qualify as sufficiently supported
 2) refactor some more to have this option specifically for json portion - probably json will have its own basic types
 3) split off the sub libs that are independently useful - at least json and pplx. see also the isolate components PR that I built after this (which is why it is draft PR)
 4) remove the option for the moment, but keep most of the #if corrections. Then go to one of the earlier options and/or put the remaining changes in a semi-supported branch for now.

I do not think squashing is good way to go for this PR. Merge is good as (1) isolated component branch is derived from this and (2) I made targeted commits for preserving history in split files.

jason-ha:
I prepped PR #1230 if option 4 is desired. It is exactly this branch plus one change to remove CMakeLists.txt related changes and the basic_types.h use of the option.

jason-ha:
@BillyONeal been a while, but this json library is still our best option. Any changes such that we might be able to get this merged?
Note: the failures since merging master into branch are all pre-existing/unrelated to these changes. Doesn't appear to be worth requeuing.

jason-ha:
Didn't mean to request review for this one. Meant for the alternate that just handles better ifdef'ing. Per previous discussion this form would be for the semi-supported branch.

Issue #1216, opened 2019-07-31
string_view support (at least for json parsing)


jason-ha:
While there is a lot of string reading in project, C++17 string_views are not supported. Modern project that are already using string_view will incur string copy penalty to create string for use by cpprest api.

I have experimented with adding an option to support string_view. Many of "const some_string_type&" arguments can be switched over.

toughengineer:
It would be nice to add `string_view` overloads for `to_string_t` functions:
https://github.com/microsoft/cpprestsdk/blob/7fbb08c491f9c8888cc0f3d86962acb3af672772/Release/include/cpprest/asyncrt_utils.h#L103

PR #1217, opened 2019-08-01
Basic string_view support

jason-ha:
addresses issue #1216 
Add CPPREST_USE_STRING_VIEWS option to accept string_views as input
- default to enable CPPREST_USE_STRING_VIEWS when compiler supports C++ 17

most string utility and json APIs are changed to support string_views
(deprecated APIs left alone)

primary argument replacements:
const utility::string_t& -> utility::string_view_t
const std::string& -> utility::nstring_view_t
const std::wstring& -> utility::wstring_view_t
const utf16string& -> utf16string_view

potential client fallout when enabling option:
For clients passing char* to conversion that does not require conversion, string_view will be the result. If string type is actually required then an explicit ctor would need to be added. Otherwise client may be able to use more string_view in own code and futher reduce memory copies. See searchfile.cpp in Samples for example.

Use of foo.c_str() and &foo[0] on a potential string_view is now foo.data().

This PR also includes changes from PR #1199 
To see changes with that PR as baseline review [jason-ha#2](https://github.com/jason-ha/cpprestsdk/pull/2).


Release/include/cpprest/string_utils.h
```diff
@@ -0,0 +1,441 @@
...
+/// Converts to a platform dependent Unicode string type.
+/// </summary>
+/// <param name="s">A single byte character UTF-8 string.</param>
+/// <returns>A platform dependent string type.</returns>
+#ifdef _UTF16_STRINGS
+_ASYNCRTIMP utility::string_t __cdecl to_string_t(utility::nstring_view_t s);
+#else
+inline const utility::string_t& to_string_t(const std::string& s) { return s; }
+#if CPPREST_USE_STRING_VIEWS
+inline utility::string_view_t to_string_t(utility::nstring_view_t s) { return s; }
```
toughengineer, line 106: All overloads of `to_string_t` should return a string object, not a string view, since the expectations of the function is to return a string object:
```
/// <param name="s">A single byte character UTF-8 string.</param>
```

If you sometimes return a string object and sometimes a string view it may lead to unexpected disasters, e.g.:
```
auto text = to_string_t(getSomeText());
```
If `getSomeText()` returns a string object or a string view, you're OK.
But if it returns something convertible to string view, after that statement the returned temporary is destroyed, `text` is dangling now and you're most certainly screwed.

PR #1276, opened 2019-11-22
Added comparison overrides to utility::datetime

jsaylor525:
I was trying to compare timestamps and noticed that the code doesn't support this so std::max and std::min don't work with datetime timestamps. I've added less/greater/less-or-equal/greater-or-equal operators.


jsaylor525:
The failure is with a bash for MAC, doesn't seem to be related to my changes...

PR #1325, opened 2020-02-10
Use websocketpp from hunter

john-forrest:
Version previously setup as hunter patch.

Look for websocketpp in hunter and use that rather than repo
version - piggy-back on existing mechanism that looks for an installed
version of websocketpp and tries to use that.


john-forrest:
Wrong repo - closing.

PR #1344, opened 2020-02-24
libcpprestsdk: fix building as a static library

ffontaine:
On Unix, try to use pkg-config to find OpenSSL. This will automatically
find any dependent libraries and put them in the correct order for
linking. If pkg-config is not available or system is not UNIX, fallback
on current mechanism

Signed-off-by: Adam Duskett <aduskett@gmail.com>
[Retrieved (and slightly updated) from:
https://git.buildroot.net/buildroot/tree/package/libcpprestsdk/0001-libcpprestsdk-fix-building-as-a-static-library.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


ffontaine:
This is a new tentative following https://github.com/microsoft/cpprestsdk/pull/1026, I'll try to fix all issues this time to get it merged.

c72578:
@ffontaine could you please test, if building as a static library works in your case, if `INTERFACE` is used instead of `PRIVATE`:
```cmake
target_link_libraries(cpprest INTERFACE ${OPENSSL_LDFLAGS})
```
See this issue for background info and a suggested patch:
https://github.com/microsoft/cpprestsdk/issues/1388#issuecomment-619594577

An alternative would be to use the keyword `PUBLIC`.
> Libraries and targets following PUBLIC are linked to, and are made part of the link interface.
https://cmake.org/cmake/help/latest/command/target_link_libraries.html

c72578:
@ffontaine could you please describe in details, under which conditions "building as a static library" fails, so that it can be reproduced? Including OS and versions of involved packages. Which error do you get?

So far, I do not see any problems, when building cpprestsdk as a static library:
```console
cd Release
mkdir build
cd build
cmake -G Ninja .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=0
ninja
cd Binaries
./test_runner
```
OK, no problems detected

Tested under:
```
OS: Linux, Fedora 31 x86_64
cpprestsdk: v2.10.15 (which is just before the commit from this PR here)
cmake: 3.17.1
openssl: 1.1.1d
pkgconf: 1.6.3
```

c72578:
@aduskett could you please describe in detail, under which conditions "building as a static library" fails, so that it can be reproduced? Including OS and versions of involved packages. Which error do you get?

ffontaine:
Build failure was raised on buildroot with version 2.10.1 using BR2_STATIC_LIBS=y which means that **all packages including openssl will be built statically by buildroot**, full debug log can be found here: http://autobuild.buildroot.net/results/be9e8d1717968a0ff8f01f7fadfa79825ac88b94/build-end.log, especially:

```
[ 89%] Linking CXX executable ../../../Binaries/blackjackclient
/accts/mlweber1/instance-2/output/host/mips64el-buildroot-linux-uclibc/sysroot/usr/lib/libcrypto.a(c_zlib.o): In function `zlib_stateful_expand_block':
c_zlib.c:(.text+0x7c): undefined reference to `inflate'
c_zlib.c:(.text+0x84): undefined reference to `inflate'
/accts/mlweber1/instance-2/output/host/mips64el-buildroot-linux-uclibc/sysroot/usr/lib/libcrypto.a(c_zlib.o): In function `zlib_stateful_compress_block':
c_zlib.c:(.text+0x13c): undefined reference to `deflate'
c_zlib.c:(.text+0x144): undefined reference to `deflate'
```

I'm able to reproduce he build failure locally with version 2.10.5 (without this patch).

I assume that if you're building on Fedora or any other "standard" OS, you'll not encounter this issue as the libssl-dev package will **always install both libcrypto.so and libcrypto.a**.

Issue #1349, opened 2020-02-29
websocket_client and json:value are oil and water on Windows


DHowett-MSFT:
* `websocket_client` allows you to receive message bodies as streams of uint8_t or as `std::string`s, which are both utf-8 (also, per specification, the only valid encoding for JSON is utf-8)
* `json::value::parse` allows you to pass `utility::string_t` or `utility::istream_t`, which on Windows are typedef'd to `wstring` and `basic_stream<wchar_t>` (`wistream`) (also, per specification, the only valid encoding for JSON is utf-8)

Without jumping through hoops, it is practically impossible to receive a message from a websocket client and parse it as JSON.

garethsb-sony:
There are `std::istream` (UTF-8) overloads on Windows. It would make sense to me to add `std::string` overloads as well.

https://github.com/microsoft/cpprestsdk/blob/b94bc32ff84e815ba44c567f6fe4af5f5f6b3048/Release/include/cpprest/json.h#L439-L459

garethsb-sony:
Resolved by #1350 (in 2.10.16)?

PR #1350, opened 2020-03-02
Expose json::value::parse for UTF8 string on Windows

garethsb-sony:
Surfactant for @DHowett-MSFT's #1349.

More functionality with less code!


garethsb-sony:
The few test failures are unrelated.

Issue #593, opened 2017-11-05
Readme issue - CMake version 


garethsb-sony:
Readme says minimum CMake version 3.7 but the cpprestsdk-config.cmake uses find_dependency macro with COMPONENTS arguments for Boost which is only supported from 3.9:

""

garethsb-sony:
fc1f692dbeb639c5eff8f31292dc5825f26c787c suggests requirement for 3.9 was not intentional...

PR #1383, opened 2020-04-05
CMake fixes + CMake search for OpenSSL (macOS)

Clovel:
- Homebrew sometimes appends the version of a package to it' directory name. Thus, one could have `/usr/local/Cellar/openssl@1.1/` instead of `/usr/local/Cellar/openssl/`, or even both. I fixed the search for `OPENSSL_ROOT_DIR` and limited it's contents to one path only. 
- Added the `<package>-config-version.in.cmake` file configuration for CMake `find_package` usage. 
- Corrected condition in `<package>-config.in.cmake` for Boost usage. There was a `@CPPREST_USES_BOOST@ AND OFF` condition, which is always false. The result was that the Boost libraries and includes were not included in the users CMake project. 
- Use `CMAKE_INSTALL_INCLUDEDIR` to install headers. `cpprestsdk` already uses `CMAKE_INSTALL_LIBDIR` & `CMAKE_INSTALL_BINDIR`


garethsb-sony:
> * Corrected condition in `<package>-config.in.cmake` for Boost usage. There was a `@CPPREST_USES_BOOST@ AND OFF` condition, which is always false. The result was that the Boost libraries and includes were not included in the users CMake project.

See this commit and the comment upon it... https://github.com/microsoft/cpprestsdk/commit/fc1f692dbeb639c5eff8f31292dc5825f26c787c

Clovel:
> > * Corrected condition in `<package>-config.in.cmake` for Boost usage. There was a `@CPPREST_USES_BOOST@ AND OFF` condition, which is always false. The result was that the Boost libraries and includes were not included in the users CMake project.
> 
> See this commit and the comment upon it... [fc1f692](https://github.com/microsoft/cpprestsdk/commit/fc1f692dbeb639c5eff8f31292dc5825f26c787c)

Comments on a lone commits will be lost in the mass of the project. I still commented. 
>People need this to use find_package(cpprestsdk). If this stays in master, we must change the CMake instructions to add Boost requirements.

And

> And anyway :
>
>1. If this causes problems with versions of CMake before 3.9, then add that as a condition instead.
>2. If cpprestsdk requires Boost, then we must keep Boost as a CMake requirement. If there are issues, then the project must be fixed.
This repair only hides the problem. And the cpprestsdk instructions are to up to date to take this into account. A work around must be described for users of this project.

Clovel:
I don't have time to look into the whole issue. I tried to use `cpprestsdk` for one of my projects. If this repository's project doesn't work as intended, I won't use it. CMake is a widely used build tool. A project like `cpprestsdk` should have a working CMake package mechanism. 

It should also have a GitHub Action for macOS.

garethsb-sony:
I agree with you, I commented because in the time I invested in reviewing your changes I found that (lost) background.

Clovel:
> I agree with you, I commented because in the time I invested in reviewing your changes I found that (lost) background.

Somebody needs to find what the cause of the issue is, so that this can be patched. 

Thanks for researching the background of this change.

garethsb-sony:
> * Added the `<package>-config-version.in.cmake` file configuration for CMake `find_package` usage.

I've been missing having the version info available for a long time. The `find_package` docs seem to indicate there are more variables we should/could set. I wonder if we can use [`write_basic_package_version_file`](https://cmake.org/cmake/help/v3.17/module/CMakePackageConfigHelpers.html#generating-a-package-version-file) to do it?

Clovel:
I think the result is the same. My solution is inspired by the files I found on my computer for the Qt CMake Version files. I took a look at other projects that implement this the way I did. 

If you believe there is another, better way to do this, I’ll take a look. 

However, the link you gave seems to describe a relatively similar method to generate the configuration file.

garethsb-sony:
> I think the result is the same. 

Ah yes, sorry, I realise your code must ultimately be derived from same root as current CMake config version file template for `COMPATIBILITY AnyNewerVersion` - https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/BasicConfigVersion-AnyNewerVersion.cmake.in

I'm not sure on C++ REST SDK backward compatibility policy between major/minor releases but maybe `SameMajorVersion` is appropriate for it?

Clovel:
> I'm not sure on C++ REST SDK backward compatibility policy between major/minor releases but maybe SameMajorVersion is appropriate for it?

Maybe. I'm not familiar enough with the project to know, sorry.

c72578:
`COMPATIBILITY AnyNewerVersion` is often preferred in upstream projects and causes less (sometimes needless) downstream issues. The primary goal is usually to just provide the version for cmake and not to restrict things. Downstream projects, which use cpprestsdk can still implement a version check, if they need to.

garethsb-sony:
> > I agree with you, I commented because in the time I invested in reviewing your changes I found that (lost) background.
> 
> Somebody needs to find what the cause of the issue is, so that this can be patched.
> 
> Thanks for researching the background of this change.

See #593 - the issue is the use of `COMPONENTS` in `find_dependency` which wasn't supported before CMake 3.9. I agree it would be better to bump minimum required version to <=3.9 and delete the `AND OFF`.

@Clovel, how do you think about bumping `cmake_minimum_required` to 3.9 in this PR? It should be done in **Release/CMakeLists.txt** of course, but preferably also (root) **CMakeLists.txt**, **README.md** (example) and there's also **Build_iOS/CMakeLists.txt**.

👍

Clovel:
CMake 3.9 is a rather recent but well used version. I personally have used more recent version on older projects with success. I think we could use this without problem. 

I’ll revise my PR to add these changes.

Clovel:
> @Clovel, how do you think about bumping `cmake_minimum_required` to 3.9 in this PR? It should be done in **Release/CMakeLists.txt** of course, but preferably also (root) **CMakeLists.txt**, **README.md** (example) and there's also **Build_iOS/CMakeLists.txt**.

Done.

garethsb-sony:
Great. What do you think about updating the `cmake_minimum_required` in the **README.md** as well?

We need @BillyONeal's review now. :-)

Clovel:
> Great. What do you think about updating the `cmake_minimum_required` in the **README.md** as well?

I missed that. Fixed it.

BillyONeal:
If you're trying to fix macos can you turn that back on in azure-pipelines.yml?

garethsb-sony:
FWIW, Homebrew was excised in c81367133c585f046b11858e07a8536b7e0611d4.

Clovel:
> If you're trying to fix macos can you turn that back on in azure-pipelines.yml?

Done.

BillyONeal:
#1392 might fix it

garethsb-sony:
I was about to say I couldn't see where the submodule was checked out! Thanks, @BillyONeal!

azure-pipelines.yml
```diff
@@ -68,93 +68,97 @@ jobs:
...
+        mkdir build.release
+      displayName: Make Build Directories
+    - task: CMake@1
+      inputs:
+        workingDirectory: 'build.debug'
+        cmakeArgs: '-G Ninja -DCMAKE_MAKE_PROGRAM=../vcpkg/downloads/tools/ninja-1.10.0-osx/ninja.exe -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake ..'
+    - task: CMake@1
+      inputs:
+        workingDirectory: 'build.release'
+        cmakeArgs: '-G Ninja -DCMAKE_MAKE_PROGRAM=../vcpkg/downloads/tools/ninja-1.10.0-osx/ninja.exe -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake ..'
```
c72578, line 137: ninja-1.10.0-osx/ninja.exe->ninja-1.10.0-osx/ninja

azure-pipelines.yml
```diff
@@ -68,93 +68,97 @@ jobs:
...
+    - script: |
+        cd build.debug
+        ../vcpkg/downloads/tools/ninja-1.10.0-osx/ninja.exe
```
c72578, line 140: ninja-1.10.0-osx/ninja.exe->ninja-1.10.0-osx/ninja

azure-pipelines.yml
```diff
@@ -68,93 +68,97 @@ jobs:
...
+      displayName: 'Run ninja debug'
+    - script: |
+        cd build.debug/Release/Binaries
+        ./test_runner *test.dylib
+      displayName: 'Run Tests debug'
+    - script: |
+        cd build.release
+        ../vcpkg/downloads/tools/ninja-1.10.0-osx/ninja.exe
```
c72578, line 148: ninja-1.10.0-osx/ninja.exe->ninja-1.10.0-osx/ninja

azure-pipelines.yml
```diff
@@ -68,93 +68,97 @@ jobs:
...
+        ./vcpkg/vcpkg install zlib openssl boost-system boost-date-time boost-regex websocketpp boost-thread boost-filesystem boost-random boost-chrono boost-interprocess brotli --vcpkg-root ./vcpkg
+      displayName: Vcpkg install dependencies
+    - script: |
+        mkdir build.debug
+        mkdir build.release
+      displayName: Make Build Directories
+    - task: CMake@1
+      inputs:
+        workingDirectory: 'build.debug'
+        cmakeArgs: '-G Ninja -DCMAKE_MAKE_PROGRAM=../vcpkg/downloads/tools/ninja-1.10.0-osx/ninja.exe -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake ..'
```
c72578, line 133: ninja-1.10.0-osx/ninja.exe->ninja-1.10.0-osx/ninja

BillyONeal:
Megaderp :)

BillyONeal:
@Clovel iOS is still toast. Do you want to investigate that or just disable it?

BillyONeal:
> The .exe extension should be removed from ninja.exe in case of osx

Fixed!

Clovel:
> @Clovel iOS is still toast. Do you want to investigate that or just disable it?

I’ll take a look in the coming days. If I don’t feel it, I’ll disable the build plan.

BillyONeal:
> > @Clovel iOS is still toast. Do you want to investigate that or just disable it?
> 
> I’ll take a look in the coming days. If I don’t feel it, I’ll disable the build plan.

This passed so I'll merge and do a release, if you end up fixing it that'd be much appreciated :)

BillyONeal:
Thanks for your contribution!

PR #1393, opened 2020-04-22
Fix uncaught exception

garethsb-sony:
PR #1310 added TCP_NODELAY as a performance optimization to ASIO-based `http_client` and `http_listener`. Unfortunately, the innocent-looking `set_option` can throw `boost::system::system_error`, which was not handled and therefore results in `std::terminate`.

I've never seen this happen on various Windows and Linux environments, but on a macOS VM we have now seen the `http_client`-side issue if the server closes the socket with just the right timing. It's hard to test, often takes 10 minutes in that environment.

My approach here is to ignore errors (we've seen EBADF and EINVAL), since this is a performance optimization only, and hard errors will be picked up by the write operation on the socket which follows immediately.

### DIFF
--- a/Release/src/http/client/http_client_asio.cpp
+++ b/Release/src/http/client/http_client_asio.cpp
@@ -333,7 +333,8 @@ public:
     void enable_no_delay()
     {
         boost::asio::ip::tcp::no_delay option(true);
-        m_socket.set_option(option);
+        boost::system::error_code error_ignored;
+        m_socket.set_option(option, error_ignored);
     }
 
 private:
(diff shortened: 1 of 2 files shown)


reneme:
Wow!

Just a few days ago, we got reports from our QA team about crashes in our latest release candidate (based on cpprestsdk 2.15.0). They are testing in a macOS VMs (!). Neither on the macOS development machines nor on Windows VMs of the QA team did we manage to reproduce those crashes.

We just started looking into that issue this morning and didn't have a clue what was going on so far. There's a good chance that you just saved our release!

Thank you very much!

garethsb-sony:
@reneme, I'm sorry that I broke it in the first place! Interesting to hear that you had exactly the same platform results as us.

PR #1435, opened 2020-05-29
GitHub Actions CI workflow for iOS and Android

solodon4:
This pull request adds support for GitHub Actions CI workflow for cpprestsdk on iOS and Android. GitHub Actions is free for open source repositories, so this should not cost anything, but lets us regularly validate mobile targets. Binaries produced during the build are only published on tagged commits since they take about 345M for both platforms. Android build takes about 17 min and iOS build takes about 22 min.

This change updates commit SHAs of Boost-for-Android and Apple-Boost-BuildScript repos to latest ones as well as points back to the original Boost-for-Android repo, which now includes support for 21.1 NDK available on GitHub Actions runner, which the workaround does not yet support.

I had to change 2 cmake settings for iOS. One is related to a problem with some standard headers inclusion when targeting iOS that many developers seem to be facing, which I describe in greater details in [this StackOveflow post](https://stackoverflow.com/questions/58628377/catalina-c-using-cmath-headers-yield-error-no-member-named-signbit-in-th/59780819#59780819). The other is a command-line warning that for some reason I was unable to disable the usual way via -Wno- flags.


solodon4:
@BillyONeal, have you had a chance to look at this?

solodon4:
@BillyONeal, I've just updated this to build with the latest Boost-for-Android (only SHA change, needed because GitHub runners now use 21.3 NDK) and Apple-Boost-BuildScript (their script now builds the new .xcframework, which coincidentally simplifies our script). Please have a look to save me from the misery of maintaining a slew of cpprestsdk patches in my repo.

PR #1438, opened 2020-06-01
Don't use pkg-config to detect OpenSSL if the vcpkg toolchain is detected

orudge:
When using vcpkg to build C++ REST SDK on Linux, I noticed the executables being generated had redundant references to libssl.so.1.1 and libcrypto.so.1.1. This is despite the vcpkg port for cpprestsdk including OpenSSL as a dependency, which is then built as a static library.

I tracked the issue down to the below definition in the cpprest_find_openssl.cmake file. It was defaulting to using pkg-config to find OpenSSL, which was finding the system library and linking in -lssl and -lcrypto. Adding in a check for VCPKG_TOOLCHAIN solved this issue for me - the find_package(OpenSSL) line then identifies the vcpkg-installed version of OpenSSL correctly. Builds not using vcpkg should be unaffected.


orudge:
Sorry, it looks like this can be fixed in vcpkg directly.

PR #1494, opened 2020-09-02
Don't find websocketpp if not CPPREST_EXCLUDE_WEBSOCKETS.

sighingnow:
Otherwise cmake will requires cpprestsdk available even when we not need it (and have excluded it).

### DIFF
--- a/Release/src/CMakeLists.txt
+++ b/Release/src/CMakeLists.txt
@@ -65,8 +65,10 @@ elseif(CPPREST_WEBSOCKETS_IMPL STREQUAL "wspp")
     websockets/client/ws_client_impl.h
     websockets/client/ws_client_wspp.cpp
   )
-  cpprest_find_websocketpp()
-  target_link_libraries(cpprest PRIVATE cpprestsdk_websocketpp_internal)
+  if(NOT CPPREST_EXCLUDE_WEBSOCKETS)
+    cpprest_find_websocketpp()
+    target_link_libraries(cpprest PRIVATE cpprestsdk_websocketpp_internal)
+  endif()
   cpprest_find_boost()
   cpprest_find_openssl()
   target_link_libraries(cpprest PUBLIC cpprestsdk_boost_internal cpprestsdk_openssl_internal)


sighingnow:
Hi devs, any comment on this PR ?

/cc @BillyONeal @garethsb-sony

barcharcraz:
It's in my queue, it looks pretty good to me at first glance

Release/src/CMakeLists.txt
```diff
@@ -65,8 +65,10 @@ elseif(CPPREST_WEBSOCKETS_IMPL STREQUAL "wspp")
     websockets/client/ws_client_impl.h
     websockets/client/ws_client_wspp.cpp
   )
-  cpprest_find_websocketpp()
-  target_link_libraries(cpprest PRIVATE cpprestsdk_websocketpp_internal)
+  if(NOT CPPREST_EXCLUDE_WEBSOCKETS)
```
barcharcraz, line 68: is CPPREST_EXCLUDE_WEBSOCKETS ever actually set anywhere. From my reading this will result in no actual change, since CPPREST_EXCLUDE_WEBSOCKETS will expand to "" (and cause a warning) or "false" and thus websockets will be found

I could be missing something.

barcharcraz (changes requested):

barcharcraz:
how does this differ from just setting "CPPREST_WEBSOCKETS_IMPL" to "none"?

Release/src/CMakeLists.txt
barcharcraz: I suppose one could set it on the command line, but in that case it should _really_ be an `option`

And given we already have CPPREST_WEBSOCKETS_IMPL I don't see what the point is.

sighingnow: Thanks for pointing out the usage of `CPPREST_WEBSOCKETS_IMPL` and after reading the code about `CPPREST_WEBSOCKETS_IMPL=none` branch again I have realized that this PR is unncessary.

I will close this PR, and many thanks for your review!

Issue #1210, opened 2019-07-30
OAuth2 Support Client Credentials Grant 


ursprateek27:

ursprateek27:
Reference:
https://tools.ietf.org/html/rfc6749#section-4.4

PR #1496, opened 2020-09-10
Support for oauth2 with "client_credentials" grant type 

icherniukh:
Adds API for requesting oauth2.0 access token using client credentials as specified in [The OAuth 2.0 Authorization Framework specification](https://tools.ietf.org/html/rfc6749#section-4.4)

Addresses #1210


icherniukh:
@barcharcraz could you take a look at this PR please? It's a really simple change that doesn't affect existing scenarios...
I would really appreciate it. Thanks!

barcharcraz (approved):
Looks good, thanks for including testing

PR #1503, opened 2020-09-24
Feature: Connection Timeout

reneme:
This introduces a "connection timeout" to distinguish between an "overall timeout" and a timeout for the "network connection establishment" only. To keep the API backward compatible (and also to adhere to the "principle of least astonishment") the "connection timeout" is defaulted to whatever the "global timeout" is set to. Only if a user of the library specifically sets the "connection timeout" it is actually honoured.

### Todo

[ ] implement it for the ASIO http client


Release/src/http/client/http_client_asio.cpp
```diff
@@ -579,8 +580,6 @@ class asio_context final : public request_context, public std::enable_shared_fro
 
             request_stream << CRLF;
 
-            m_context->m_timer.start();
-
```
reneme, line 583: This call moved to the call site of the surrounding method (`ssl_proxy_tunnel::start_proxy_connect()`) i.e. `asio_context::start_request()`. This is the only location where `ssl_proxy_tunnel::start_proxy_connect()` is called if a proxy-tunnel is required. `asio_context::start_request()` also initiates the ordinary connection (without a proxy-tunnel). Therefore, I found it more consistent to move the call to `start_connecting()` there.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -864,12 +863,6 @@ class asio_context final : public request_context, public std::enable_shared_fro
             // Enforce HTTP connection keep alive (even for the old HTTP/1.0 protocol).
             request_stream << "Connection: Keep-Alive\r\n\r\n";
 
-            // Start connection timeout timer.
-            if (!ctx->m_timer.has_started())
-            {
-                ctx->m_timer.start();
-            }
-
```
reneme, line 872: That is not needed anymore, as the conditional below should clearly distinguish the states of the timer:

a) The connection is reused (or an established proxy tunnel): In both cases the call to `ctx->write_request()` will eventually call `m_timer.start_communicating()`. Note that it is legal to call `start_communicating()` without a prior call to `start_connecting()` (e.g. when reusing an already established connection).

b) Otherwise, no connection is established yet, and a call to `m_timer.start_connecting()` is necessary.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -917,6 +912,8 @@ class asio_context final : public request_context, public std::enable_shared_fro
             // 'start_http_request_flow'
             std::shared_ptr<ssl_proxy_tunnel> ssl_tunnel =
                 std::make_shared<ssl_proxy_tunnel>(shared_from_this(), start_http_request_flow);
+
+            m_timer.start_connecting();
```
reneme, line 916: This replaces the above-mentioned call inside `ssl_proxy_tunnel::start_proxy_connect()`. All calls to `m_timer.start_connection()` are now consistently placed inside `asio_context::start_request()`.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -1091,6 +1088,7 @@ class asio_context final : public request_context, public std::enable_shared_fro
         }
         else
         {
+            m_timer.start_communicating();
```
reneme, line 1091: This is inside the `write_request()` that starts sending the request to the server. Just before the respective `async_write()` we call `m_timer.start_communicating()` to switch the `timeout_timer` state from the "connection timeout" to the "global timeout".

Note that this method might also initiate a TLS handshake if required (see conditional above). In that case, the request is eventually sent inside `handle_handshake()` and _not_ here.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -1101,6 +1099,7 @@ class asio_context final : public request_context, public std::enable_shared_fro
     {
         if (!ec)
         {
+            m_timer.start_communicating();
```
reneme, line 1102: As stated above, if a TLS handshake was required, this will now eventually send the request. As before, `m_timer.start_communicating()` switches the `timeout_timer` over to the "global timeout", just before sending the application payload data.

Release/src/http/client/http_client_asio.cpp
```diff
@@ -1920,15 +1915,34 @@ class asio_context final : public request_context, public std::enable_shared_fro
...
         };
 
+        void schedule_timeout()
+        {
+            const auto duration = (m_state == connecting)
+                                    ? m_connection_timeout_duration
+                                    : m_timeout_duration;
+
+            const auto cancelled_handlers = m_timer.expires_from_now(duration);
+            if (cancelled_handlers > 0 || m_state == connecting) {
```
reneme, line 1931: Note to self: This might actually fail to register a timeout handler if the connection was reused. In that case, the initial registration would happen with `m_state == communicating`.

PR #1551, opened 2020-12-31
mint 2.10.17

barcharcraz:


c72578:
Is it intentional, that the release tag is without a "v" now?
2.10.17 compared to previously v2.10.16 etc.
https://github.com/microsoft/cpprestsdk/releases

This requires a change in the download links for packaging (removing the "v"):

https://github.com/microsoft/cpprestsdk/archive/2.10.17.tar.gz
https://github.com/microsoft/cpprestsdk/archive/v2.10.16.tar.gz

PR #1558, opened 2021-01-04
Fix type mismatch in basic_file_buffer

andreas-schwab:


barcharcraz:
is this simply to silence a warning?

andreas-schwab:
No, it fixes a bug.

PR #1576, opened 2021-02-17
export http_exception for non Windows builds

JvdGlind:
If a project compiled with clang for macos links to CppRest, and the
-fvisibility=hidden compiler option is used, the http_exception is used
as std::exception. The clang compiler requires it to be exported in
order to know about the symbols.

Since the Visual Studio compiler throws warnings regarding the export of
a class that inherits from a non-exported class, std::exception in this
case, the WIN32 define check is added to omit this export.


JvdGlind:
It seems that my patch didn't work, as the dllimport is ignored on clang and I 'fixed' it by throwing the exception in the test, rather than let it originate from the library. Closing this PR until I have a better solution.

Issue #467, opened 2017-06-08
http_client sends out request sent *after* connection timeout ?!?


blavov:
Hello all,
I've got a problem with the http_client (cpprestsdk 2.9.0, downloaded from the at the beginning of February, running on Ubuntu 14.04.) making http_requests that have already timed out. This results in the operation failing on the client end, but succeeding on the web service, which is a disaster. 

I have a server where I want to add to a value to an account that can be reached via a REST/json call (PUT to a specific URL with the value in a json-formatted body); the connection timeout is set to 20 seconds. 

This work well under normal conditions, but if the network connection is down (I test this by simply unplugging the ethernet connection to the router on the Linux pc side), then things start going awry. In my test, the timeout is set to 20 seconds (via http_client_config::set_timeout() ), which results in an error displayed on screen when reached (the http_exception I get is "failed to read HTTP status line"). 

I then replug the connection, wait for the net to come up again, and press "retry" on my display, which calls the same function again. The second try works fine, but upon closer inspection, I see that the account was incremented **twice**! 

Using a packet logger, I can see that the first try is executed immediately and doesn't come very far a the network connection is down. As soon as I replug, however, that very same connection - the same source and destination port and address - retries sending out the packets, this time succeeding. But the http_client.request() already failed five seconds before I replugged the cable, and at the time when those packets go out, I have yet to press the "retry" button, so I haven't even started the manual retry! 

The client call looks like this: 

```
myClient::DoPOST()
    {
    json::value requestBody;
    http_request webRequest(methods::POST);
    webRequest.headers().add("Accept","application/json");
    webRequest.headers().add("Language","de");
    webRequest.set_request_uri(SOME_URL);
    // fill the request_body with the appropriate data
    webRequest.set_body(requestBody);
    try {
        pplx::task<http_response> webTask = webRequest.request(httpReq);
        webTask.wait();
        // now that request is over, look at response
        http_response myResponse = webTask.get();
        pplx::task<http_response> finishedResponse = myResponse.content_ready();
        finishedResponse.wait();
        // more result handling
        ... 
    } catch (http_exception &exc) {
        // some error logging
    }
```

As said above, this routine catches an http_exception "failed to read HTTP status line" on first try and then terminates - but still seems to execute the request a few seconds **after** the failure and termination. 

Is this normal behaviour for the cpprest lib? How can I prevent this? Should I do anything differently? Any help is appreciated, as I don't know how to handle this very strange phenomenon. 

[...]

PR #1581, opened 2021-02-24
disable-socket-send-buffer option added

baltendorf:
This PR adds an option to disable the underlying sockets send-buffer by setting SO_SNDBUF to zero.

During sending requests with larger bodies we encountered, especially on low bandwidth connections, that after the progress was set to 100% theire were still outgoing data to process. So visually, by following the progress, sending the data was done but technically the data was just pumped into the send-buffer of the socket and sits theire to wait for transmission.
This leads to a significant gap on low bandwidth connections between signaling 100% progress and finishing the task.
At 64 kbps and a one MB file this can be up to 90 seconds.
The same applies for connections suffering on a large error rate resulting in many retransmissions. The data was buffered just fine quick but it took a significant long time for finally closing the task.
On higher bandwidth or errorless connections the buffering take a minor part. 

This way we decide to send the data unbuffered. So the write-handler will ever been called if the data was successfully send by the socket and not if the data was just putted into the buffer. Regardless the network bandwidth or error-rate the task was finished right after the last chunk of data was send.

Thats maybe a worthwhile addition for people transferring larger data chunks and want to rely on a accurate progress and task completion state.

### DIFF
--- a/Release/include/cpprest/http_client.h
+++ b/Release/include/cpprest/http_client.h
@@ -97,6 +97,7 @@ public:
         : m_guarantee_order(false)
         , m_timeout(std::chrono::seconds(30))
         , m_chunksize(0)
+        , m_socket_send_buffer_disabled(false)
         , m_request_compressed(false)
 #if !defined(__cplusplus_winrt)
         , m_validate_certificates(true)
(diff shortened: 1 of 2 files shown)


hawkeyese:
Do you think this change could affect https://github.com/Microsoft/cpprestsdk/issues/467 as well?

PR #1624, opened 2021-05-17
Improving OpenSSL search for APPLE platform

mayanksahu:
In MacOS BigSurr while compiling the repo we are not able to find
OpenSSL (required version >= 1.0.0)
On investigation, it was happening because of two cmake variables not being set properly: OPENSSL_INCLUDE_DIR, OPENSSL_LIBRARIES

when running pkg_search_module(OPENSSL openssl) to get info on openssl, we end up with

-- OPENSSL_INCLUDE_DIR = 
-- OPENSSL_LIBRARIES = ssl;crypto
find_package(OpenSSL 1.0.0 REQUIRED), which is a fallback when pkg_search_module fails, is able to set the values properly.

-- OPENSSL_INCLUDE_DIR = /usr/local/opt/openssl@1.1/include
-- OPENSSL_LIBRARIES = /usr/local/opt/openssl@1.1/lib/libssl.dylib;/usr/local/opt/openssl@1.1/lib/libcrypto.dylib
So I'm making changes to strengthen the condition which makes us to go to fallback.


guoxiangCN:
I also had this problem， hope you pr helpful

mayanksahu:
can we merge this?

PR #1636, opened 2021-07-21
Large file support on 32bit platforms

tcoserea-msft:
- on 32bit platforms size_t is not enough to support files with size
greater than 4GB.

### DIFF
--- a/Release/include/cpprest/details/fileio.h
+++ b/Release/include/cpprest/details/fileio.h
@@ -54,8 +54,8 @@ struct _file_info
 
     // Positional data
 
-    size_t m_rdpos;
-    size_t m_wrpos;
+    utility::size64_t m_rdpos;
+    utility::size64_t m_wrpos;
     bool m_atend;
 
     // Input buffer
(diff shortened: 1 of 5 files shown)


Release/include/cpprest/details/fileio.h
```diff
@@ -54,16 +54,16 @@ struct _file_info
...
+    utility::size64_t m_wrpos;
     bool m_atend;
 
     // Input buffer
 
     size_t m_buffer_size; // The intended size of the buffer to read into.
     char* m_buffer;
 
-    size_t m_bufoff;                          // File position that the start of the buffer represents.
+    utility::size64_t m_bufoff;               // File position that the start of the buffer represents.
```
barcharcraz, line 66: This is wrong, and introduces several out of bounds reads by memcpy throughout the rest of the code

barcharcraz:
I don't have a lot of time to review this for correctness on all our platforms and so would like tests, in particular for every data value widened from size_t to int64_t I'd like to see a test that results in that data value actually exceeding UINT32_MAX. I already spotted several instances where you're using int64_t as an offset into an array, or passing it to memcpy.

I'd suggest using truncate(on linux) or SetEndOfFile (on windows) to create a very large file, and then write some known data to a few pages in the very large region.

Release/src/streams/fileio_posix.cpp
```diff
@@ -468,7 +468,7 @@ size_t _getn_fsb(Concurrency::streams::details::_file_info* info,
         if (static_cast<int>(read) > 0)
         {
             auto copy = (std::min)(read, byteCount);
-            auto bufoff = fInfo->m_rdpos - fInfo->m_bufoff;
+            auto bufoff = static_cast<size_t>(fInfo->m_rdpos - fInfo->m_bufoff);
```
barcharcraz, line 471: this is concerning to me since it seems like it could cause truncation on a size passed to memcpy, and may need a validation check.

Issue #1667, opened 2021-12-13
http_request and http_response doesn't satisfy the rule of five


evilenzo:
As you can see in https://github.com/microsoft/cpprestsdk/blob/master/Release/include/cpprest/http_msg.h http_request and http_response don't satisfy the rule of five and also have a non-trivial destructor. So no move will work on this types what is actually sad

evilenzo:
Pull Request: https://github.com/microsoft/cpprestsdk/pull/1668

PR #1668, opened 2021-12-13
Missing operators and ctors for http_request and http_response

evilenzo:
Solves this  issue: https://github.com/microsoft/cpprestsdk/issues/1667

### DIFF
--- a/Release/include/cpprest/http_msg.h
+++ b/Release/include/cpprest/http_msg.h
@@ -503,6 +503,8 @@ public:
 
     _http_response(http::status_code code) : m_status_code(code) {}
 
+    virtual _http_response() = default;
+
     http::status_code status_code() const { return m_status_code; }
 
     void set_status_code(http::status_code code) { m_status_code = code; }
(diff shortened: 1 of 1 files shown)


Release/include/cpprest/http_msg.h
```diff
@@ -548,6 +550,60 @@ class http_response
...
+        {
+            _m_impl = std::move(_Other._m_impl);
+        }
+        return *this;
+    }
+
+    /// <summary>
+    /// Destructor frees any held resources.
+    /// </summary>
+    ~http_response() = default;
```
barcharcraz, line 605: These don't appear to do anything over just accepting the compiler's defaults. The original class has an empty destructor, which is somewhat redundant, not an excuse to add even more redundant stuff

Release/include/cpprest/http_msg.h
```diff
@@ -853,7 +909,7 @@ class _http_request final : public http::details::http_msg_base, public std::ena
 
     _ASYNCRTIMP _http_request(std::unique_ptr<http::details::_http_server_context> server_context);
 
-    virtual ~_http_request() {}
+    virtual ~_http_request() = default;
```
barcharcraz, line 912: why?

Release/include/cpprest/http_msg.h
```diff
@@ -951,10 +1007,67 @@ class http_request
...
+            _m_impl = std::move(_Other._m_impl);
+        }
+        return *this;
+    }
+
     /// <summary>
     /// Destructor frees any held resources.
     /// </summary>
-    ~http_request() {}
+    ~http_request() = default;
```
barcharcraz, line 1070: again not needed

barcharcraz (changes requested):

barcharcraz:
This approach seems completely backwards, for starters it adds a bunch of (implicit) copy constructors and that's totally not necessary, and also, the new move constructor doesn't move!

It would be much better to stop defining an empty destructor (which is equivalent to what the compiler would do anyways) and just default the move constructor, so the shared pointer would get moved

evilenzo:
Ok. My point was to implement all functions of rule of five to not be aware if compiler won't generate move/copy ctors and assignment operators. If we wan't to produce less redunant code, we must follow standard which says that compiler will generate both copy and assignment functions if we doesn't  have user-defined destructor. The main problem is [this](https://github.com/microsoft/cpprestsdk/blob/bfe34877914fd280512c97d41aaa99e88564fd8d/Release/include/cpprest/http_msg.h#L957) line. User-defined destructor means that compiler won't support us

evilenzo:
Also this file has other ctors and dtors that are defaulted. If it is the matter of style so why we won't change them to {} too?

PR #1673, opened 2022-01-06
TLS 1.3 support with 10.0.22000.0 SDK

Karteek649:
Verified the http_client in Windows 11 OS, able to communicate in TLS1.3


barcharcraz:
This breaks the winhttppal build. And probably breaks builds with older versions too

bking-imprivata:
Is there an ETA for this to be merged?  It has been opened for quite some time.

PR #1682, opened 2022-02-08
Try to resend request if connection has been closed.

blacktea:
Hi, all!

**Case**:
We've got a client that sends messages to a server. I came across a problem when client failed to write request body. I found out that the reason was the connection was closed after the "keep-alive" timeout expired.
They use HTTP/1.1 and Keep-Alive connection.

**Tcpdump trace**:
image

155 - client, 159 - server.
At the beginning, the client sends a request and accepts a response from the server. Then, in 19 secs FIN packet is arrived. At 209.852367 we try to write headers. In response, we accepts RST packet. In cpprestsdk it means "Broken pipe" at https://github.com/microsoft/cpprestsdk/blob/master/Release/src/http/client/http_client_asio.cpp#L1303.

**My solution**:
There is an attempt to restore the closed connection at https://github.com/microsoft/cpprestsdk/blob/master/Release/src/http/client/http_client_asio.cpp#L1347.
I decided to use this method to handle body write error.
I managed to reproduce the error in unit-tests.

### DIFF
--- a/Release/src/http/client/http_client_asio.cpp
+++ b/Release/src/http/client/http_client_asio.cpp
@@ -219,7 +219,7 @@ public:
 
         // These errors tell if connection was closed.
         if ((boost::asio::error::eof == ec) || (boost::asio::error::connection_reset == ec) ||
-            (boost::asio::error::connection_aborted == ec))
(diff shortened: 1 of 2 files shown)


blacktea:
@barcharcraz, Could you review please?

PR #1723, opened 2022-10-07
workaround clang 14 __has_declspec_attribute restriction

### LINKED ISSUE #1710: cpprest_compat breaks cURL header when building with Clang 14
`#define dllimport` inside [cpprest_compat.h](https://github.com/microsoft/cpprestsdk/blob/80444805898842f7e5c62b256337c0897eb254bb/Release/include/cpprest/details/cpprest_compat.h) breaks [curl.h](https://github.com/curl/curl/blob/bfbde883af33397943df68a3ae01847a634d33bf/include/curl/curl.h#L120) on Clang 14.

See minimized example here: [godbolt](https://godbolt.org/z/T6nxGY4cz)

The issue was initially [reported to LLVM](https://github.com/llvm/llvm-project/issues/53269).
The LLVM issue was closed. For the reason, see these comments: [comment1](https://github.com/llvm/llvm-project/issues/53269#issuecomment-1158813473) and [comment2](https://github.com/llvm/llvm-project/issues/53269#issuecomment-1161732686).

liam-x:
clang 14 impose a new restriction on __has_declspec_attribute, the argument can not be empty.
https://github.com/llvm/llvm-project/issues/53269

cppresetsdk [cpprest_compat.h](https://github.com/microsoft/cpprestsdk/compare/master...liam-x:cpprestsdk:fix-issue-1710?expand=1#diff-42889c5f05ff275fe4036ca6b5be477e7869ce5a1a74094ba0b2f92a318b4e9e) translate dllimport to empty which cause compilation failure on clang 14.

wordaround by translate dllimport to an unrelated identifier to fix #1710 .

### DIFF
--- a/Release/include/cpprest/details/cpprest_compat.h
+++ b/Release/include/cpprest/details/cpprest_compat.h
@@ -29,7 +29,7 @@
 #else // ^^^ _WIN32 ^^^ // vvv !_WIN32 vvv
 
 #define __declspec(x) __attribute__((x))
-#define dllimport
+#define dllimport   x
 #define novtable /* no novtable equivalent */
 #define __assume(x)                                                                                                    \
     do                                                                                                                 \


Release/include/cpprest/details/cpprest_compat.h
```diff
@@ -29,7 +29,9 @@
 #else // ^^^ _WIN32 ^^^ // vvv !_WIN32 vvv
 
 #define __declspec(x) __attribute__((x))
-#define dllimport
```
barcharcraz, line 32: I don't think this is a good way to fix this. There are not that many appearances of `__declspec` in the codebase, and most are already behind visibility macros, the remaining are noinline, noreturn, and novtable. Noreturn can probably be replaced by writing [[noreturn]] directly, noinline should be replaced by something like 

```c++
#ifdef _MSC_VER
#define CPPREST_NOINLINE __declspec(noinline)
#else
#define CPPREST_NOINLINE __attribute__((noinline))
#endif
```
and then replacing `__declspec(noinline)` usages with CPPREST_NOINLINE

the usages with dllimport/dllexport should be replaced by a proper set of visibility macros, cmake can generate them but they are also easy to write, something like:

```c++
#ifdef _MSC_VER
    #define CPPREST_API_CLASS
    #if defined(CPPREST_BUILDING) && defined(CPPREST_DLL)
        #define CPPREST_API __declspec(dllexport)
    #elif defined(CPPREST_DLL)
        #define CPPREST_API __declspec(dllimport)
    #else
        #define CPPREST_API
    #endif
#else
    // it might be a good idea to do something special for static libraries here,
    // since this could result in apps exporting symbols to loaded DSOs instead of those
    // DSOs loading cpprestsdk themselves.
    #define CPPREST_API_CLASS __attribute__((visibility("default")))
    #define CPPREST_API __attribute__((visibility("default")))
    #endif
#endif
```

any classes used as exceptions (or generally if the RTTI information needs exporting) nned to be annotated as `CPPREST_API_CLASS`, this should be a seperate macro because on windows applying dllexport to a class can export all it's bases, which is extremely annoying

barcharcraz (changes requested):

PR #1734, opened 2022-11-30
Use the new MSVC preprocessor

abrauninger:
MSVC has a new preprocessor (https://devblogs.microsoft.com/cppblog/announcing-full-support-for-a-c-c-conformant-preprocessor-in-msvc/).  To compile code with MSVC and C++20 we need to use the new preprocessor (`/Zc:preprocessor`).

This change updates cpprestsdk to use `/Zc:preprocessor` on versions of MSVC that support it (>= 16.6, which is _MSC_VER 1926).

Without code changes, `/Zc:preprocessor` causes build breaks due to empty `__VA_ARGS__` in the macros defined in CheckMacros.h and TestMacros.h.  Those headers already have versions of the macros that use the conformant `##__VA_ARGS__` syntax for non-MSVC compilers, so this change uses those existing macros for MSVC as well, on versions of MSVC that support `/Zc:preprocessor`.


tiagomacarios:
Perhaps the we should enabling the new code based on [_MSVC_TRADITIONAL](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170#:~:text=set.%20Otherwise%2C%20undefined.-,_MSVC_TRADITIONAL,-%3A)

barcharcraz:
Is this really necessary? this appears to only effect internal test headers, not public API headers, So I don't think this should fix any bugs related to using cpprest and the new preprocessor at the same time.

barcharcraz (approved):

PR #1746, opened 2023-03-09
Fix overflow exception when serializing

25077667:
When the MSB of the initial wchar_t is set, the lowSurrogate would be inferred to be "unsigned".
The if statement would be triggered, and the exception "UTF-16 string has invalid low surrogate" would be thrown.

This needs to be fixed by making the type "signed" and then comparing it to the integer values of the specified constants.


25077667:
You could also refer to:
The wchar_t type is defined by MIDL as an unsigned short(16-bit) data object [1].

Type wchar_­t is a distinct type that has an implementation-defined signed or unsigned integer type as its underlying type [2].

[1] https://learn.microsoft.com/en-us/windows/win32/midl/wchar-t
[2] https://eel.is/c++draft/basic.types#basic.fundamental-8

PR #1758, opened 2023-05-19
Fix bug when offset is not applied in seekof(..., std::ios_base::end, ...)

dimhotepus:
`offset` was not applied when we seek from end.


raychaudhuri-amitava (approved):
Changes look good to me

PR #1780, opened 2023-11-23
fix issue: The web_proxy doesn't work when set proxy with uri

ceasor:
http_client_winhttp::send_request only handle the case when proxy.is_auto_discovery() return true. 
image
So when we construct a web_proxy with uri, it will only set the proxy mode as user_provided_. 
image
So send_request will ignore the proxy setting when we use uri to construct a web_proxy.

My modification is try to fix the problem.


ceasor:

PR #1824, opened 2025-04-23
[PPLX] Fixed bug in pplx::task_options(cancellation_token, task_continuation_context) constructor

eugenegff:
Fixed bug in pplx::task_options(cancellation_token, task_continuation_context) constructor - should set _M_HasCancellationToken to true. Bug was introduced in PPL in VS2013 and fixed in VS2015 but survives in PPLX. Note - it is one of 4 .ctors, bug is only in this one


eugenegff:
@StephanTLavavej FYI

StephanTLavavej:
I don't maintain cpprestsdk, sorry.

PR #1829, opened 2025-06-21
Fix missing std::char_traits<uint8_t> (if it is in fact missing). #1812

leakingmemory:


leakingmemory:

leakingmemory:
Just to clarify on this. My personal opinion is that it is better to break compatibility and fix this by not using unsigned char types with C++ streams. It has worked fine on some compilers and standard libraries, but to my knowledge only char types are supported by the C++ standard. This pull request was made with the intention to not break compatibility with existing code.

yuxiang-wen:
收到，谢谢

Issue #1212, opened 2019-07-30
STATUS_STACK_BUFFER_OVERRUN in completion_callback(WINHTTP_CALLBACK_STATUS_HEADERS_AVAILABLE) 


acs-godonnell:
https://github.com/microsoft/cpprestsdk/blob/master/Release/src/http/client/http_client_winhttp.cpp#L2061

```
case WINHTTP_CALLBACK_STATUS_HEADERS_AVAILABLE:
            {
                // First need to query to see what the headers size is.
                DWORD headerBufferLength = 0;
                query_header_length(hRequestHandle, WINHTTP_QUERY_RAW_HEADERS_CRLF, headerBufferLength);

                // Now allocate buffer for headers and query for them.
                std::vector<unsigned char> header_raw_buffer;
                header_raw_buffer.resize(headerBufferLength);
                utf16char* header_buffer = reinterpret_cast<utf16char*>(&header_raw_buffer[0]);
```

We need to check that `headerBufferLength `> 0 before accessing `header_raw_buffer[0]` here.

I'd left my application running in the debugger, the virtual machine suspended and resumed, and the resulting STATUS_STACK_BUFFER_OVERRUN (0xC0000409) took down the application.

WolfgangVogl:
I've got the STATUS_STACK_BUFFER_OVERRUN, too - i think it's this issue.
Beside of the source code you've postet i think the same problem can occur for
https://github.com/microsoft/cpprestsdk/blob/master/Release/src/http/client/http_client_winhttp.cpp#L115
`static utility::string_t parse_reason_phrase(HINTERNET request_handle)
{
    utility::string_t phrase;
    DWORD length = 0;

    query_header_length(request_handle, WINHTTP_QUERY_STATUS_TEXT, length);
    phrase.resize(length);`

PR #1837, opened 2025-11-12
parse HTTP headers without asking for them again

Slawomir2:
This fixes issue https://github.com/microsoft/cpprestsdk/issues/1212

The old code was calling WinHttpQueryHeaders 2nd time but that was sometimes resulting in headers not being found in case the request_handle was reset in the meantime

WinHttpQueryHeaders would then return false and GetLastError() would return ERROR_INVALID_HANDLE.

Eventually application would crash due to uncaught exception thrown by stoi() function when it was passed an empty string (exit code 0xc0000409 which does not necessarily mean stack overflow
- see https://devblogs.microsoft.com/oldnewthing/20190108-00/?p=100655).

### DIFF
--- a/Release/tests/common/UnitTestpp/src/TestRunner.cpp
+++ b/Release/tests/common/UnitTestpp/src/TestRunner.cpp
@@ -39,6 +39,7 @@
 #include <agents.h>
 #include <functional>
 #else
+#include <chrono>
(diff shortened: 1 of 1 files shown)


Slawomir2:
Code removed

